PR #8, opened 2020-08-14
Weiser/add prefixed docs

weiser:
I'm deploying my huma application behind a shared API gateway.  As such, I need all of my routes to be prefixed with my app name, i.e.
```
GET endpoint1 --> GET my-app/endpoint
```

For routes I myself can define, that is easy enough.

Buyt for the API docs, which have static routes of `/{docs, openapi.json,openapi.yaml}` I have no way to change those routes to
```
GET /docs --> GET /my-app/docs
```

This PR adds a new RouterOption, which when applied ensures that the auto-generated docs can be prefixed.

### DIFF
--- a/docs.go
+++ b/docs.go
@@ -19,10 +19,10 @@ func splitDocs(docs string) (title, desc string) {
 	return
 }
 
-// RapiDocHandler renders documentation using RapiDoc.
-func RapiDocHandler(pageTitle string) Handler {
-	return func(c *gin.Context) {
-		c.Data(200, "text/html", []byte(fmt.Sprintf(`<!doctype html>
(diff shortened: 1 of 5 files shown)


weiser:
Based on the result of review, I need to update the main docs to show how to use this new Router Option?

weiser:
This should not be accepted.  Though the tests pass, when I wire this up to an actual running instance, the DOM doesn't render because the `rapi-doc` element is missing the prefixed path.  I.e.

 `<rapi-doc spec-url="openapi.json"...>` instead of   `<rapi-doc spec-url="prefix/openapi.json"...>`

weiser:
closed in favor of https://github.com/danielgtaylor/huma/pull/9

PR #9, opened 2020-08-18
docs should be available on a subpath of /

weiser:
Previously, the `/docs` endpoint was static, meaning that if my huma-powered app was meant to be accessed from `/myapp/*` the docs could not be accessed.

This PR addresses that (in a pretty hacky way) by:
1. enabling a path to be prefixed to where the docs are mounted.
2. templating the rapi, redoc and swagger DOMs so that we need a title and a path to the openapi.json file.

This work should invalidate #8 .

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -418,6 +418,10 @@ func (w *contentEncodingWriter) WriteHeader(code int) {
 	w.status = code
 }
 
+func (w *contentEncodingWriter) Status() int {
+	return w.status
+}
+
 func (w *contentEncodingWriter) Close() {
 	if !w.wroteHeader {
 		w.ResponseWriter.WriteHeader(w.status)
--- a/middleware_test.go
+++ b/middleware_test.go
@@ -237,3 +237,28 @@ func TestContentEncodingCompressedMultiWrite(t *testing.T) {
 	decoded, _ := ioutil.ReadAll(gr)
 	assert.Equal(t, 2250, len(decoded))
 }
+
+func TestContentEncodingError(t *testing.T) {
+	var status int
+
+	r := NewTestRouter(t)
+	r.GinEngine().Use(ContentEncodingMiddleware())
+	r.GinEngine().Use(func(c *gin.Context) {
+		c.Next()
+
+		// Other middleware should be able to read the response status
+		status = c.Writer.Status()
+	})
+	r.GinEngine().GET("/", func(c *gin.Context) {
+		c.Writer.WriteHeader(http.StatusNotFound)
+		c.Writer.Write([]byte("some text"))
+	})
+
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest(http.MethodGet, "/", nil)
+	req.Header.Add("Accept-Encoding", "gzip")
+	r.ServeHTTP(w, req)
+
+	assert.Equal(t, http.StatusNotFound, status)
+	assert.Equal(t, http.StatusNotFound, w.Result().StatusCode)
+}


docs.go
```diff
@@ -19,10 +19,10 @@ func splitDocs(docs string) (title, desc string) {
...
 }
 
-// RapiDocHandler renders documentation using RapiDoc.
-func RapiDocHandler(pageTitle string) Handler {
-	return func(c *gin.Context) {
-		c.Data(200, "text/html", []byte(fmt.Sprintf(`<!doctype html>
+// RapiDocTemplate is the template used to generate the RapiDoc.  It needs two args to render:
+// 1. the title
+// 2. the path to the openapi.yaml file
+var RapiDocTemplate = `<!doctype html>
```
danielgtaylor, line 25: Should these be mutable? If not, we may want a func to return the string. Could also just be a func that renders the template, so you can take in args to describe the inputs instead of using comments.

options.go
```diff
@@ -369,12 +376,23 @@ func HTTPServer(server *http.Server) RouterOption {
 // DocsHandler sets the documentation rendering handler function. You can
 // use `huma.RapiDocHandler`, `huma.ReDocHandler`, `huma.SwaggerUIHandler`, or
 // provide your own (e.g. with custom auth or branding).
+//
+// DEPRECATED! Use `DocsDomType` instead!
 func DocsHandler(f Handler) RouterOption {
+	fmt.Println("This option is deprecated, use `DocsDomType` instead")
```
danielgtaylor, line 382: Can this be supported without deprecating it?

options.go
```diff
@@ -369,12 +376,23 @@ func HTTPServer(server *http.Server) RouterOption {
...
 	return &routerOption{func(r *Router) {
 		r.docsHandler = f
 	}}
 }
 
+// DocsDomType sets the presentation for the docs UI.  Valid values are:
+// "rapi" (default), "redoc", or "swagger"
```
danielgtaylor, line 389: Constants or a custom type for the inputs?

weiser: They should be constants.

options.go
weiser: Strictly speaking, yes.  Where it gets weird is if you use the `DocsRoutePrefix` RouterOption to change the prefix of the where the docs live.

E.g.
```
NewServer("api", "version", DocsRoutePrefix("/prefix"), DocsHandler(ReDocHandler(title))))
```

Since `DocsHandler` is just setting a function to call when `/docs` is hit, and `ReDocHandler` isn't aware of the new prefix, it felt weird to have both `DocsHandler`s and `DocsRoutePrefix`es.

A way to get around that would be to call:
```
NewServer("api", "version", DocsRoutePrefix("/prefix"), DocsDomType("redoc")))
```

which would give the user the ability to use ReDoc, while also ensuring that the docs are reachable at `/prefix/docs`.

Which is why I went for the deprecation route.

docs.go
weiser: Yeah, It'd be better to just export a function that would take args, instead of forcing a consumer of the module to know about the internals of a template.

I initially took the "export templates" approach because it helped me minimize the changes in all of the docs-Handlers in `docs.go`

weiser: functionized this: 2852753

options.go
weiser: constantized stuff at 1c74420

weiser:
> I also think this middleware needs to be updated:

@danielgtaylor I updated the addservicelinkmiddleware at: e1babc9. there's some stuff that could be DRY'd up, but that would come at the cost of changing the interface of the methods involved in the commit.  I learned toward not changing the interface, but understand that it'd probably be better to modify the interface (at the risk of breaking anything that depends on that method)

PR #10, opened 2020-08-20
Weiser/add prefixed docs

weiser:


router.go
```diff
@@ -374,17 +378,33 @@ func NewRouter(docs, version string, options ...RouterOption) *Router {
 		r.corsHandler(c)
 	})
 
+	// We need to ensure that if the docs have a prefixed path,
+	// that the ServiceLinkMiddleware can reflect the true path to the docs
+	r.GinEngine().Use(func(c *gin.Context) {
+		c.Set("docsPrefix", r.docsPrefix)
```
weiser, line 384: I feel pretty dirty shoving the prefix into the context, but I could see no other way to actually pass the prefix to the middleware since the middleware only takes a context.

PR #11, opened 2020-09-10
Refactor library interface

danielgtaylor:
This change is a major refactor / partial rewrite of Huma which tries to accomplish a few things:

- Simplify writing handler functions
- Support request and response streaming
- Compatibility with standard library and popular third-party libraries
  - `context.Context`
  - `http.ResponseWriter`
  - `io.Reader`
  - `io.Writer`
  - middleware `func(next http.Handler) http.Handler`
- Per-resource (aka prefixed) middleware
- Replaces complicated/confusing dependencies with resolvable input structs

See the updated `README.md` for more details.


weiser (commented):
This review is just to address things in the readme, only considering things from a consumer's POV.

Does any public API break as a result of this refactor?  If so, might we need a "upgrade from v1" section in the readme?

README.md
```diff
@@ -25,464 +24,423 @@ Features include:
...
+	for i := 0; i < 5; i++ {
+		// Use AddError to add multiple error details to the response.
+		ctx.AddError(fmt.Errorf("Error %d", i))
+	}
+
+	// Check if the context has had any errors added yet.
+	if ctx.HasError() {
+		// Use WriteError to set the actual status code, top-level message, and
+		// any additional errors. This sends the response.
+		ctx.WriteError(http.StatusBadRequest, "Bad input")
```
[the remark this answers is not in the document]
danielgtaylor: > 1. There is no explicit return OK-type of statement, which I'd expect. Do handlers default to returning OK?

Handlers don't return anything now (e.g. to support streaming), so there is no explicit `return` statement anymore. Like the standard Go library, the default response is `http.StatusOK` with an empty body if you never write anything. If you call on any of the `Write*` methods that will explicitly set the response info, e.g. `ctx.WriteHeader(http.StatusOK)`.

> 2. If I declare that a valid response is a response.RedirectPermenantly(), it isn't clear to me from this example how I'd know which http.Status* field I should use.

They all deliberately match, so `responses\.(.*)` ⇨ `http.Status$1`, e.g. `http.StatusRedirectPermanently`. The `responses` package is just a convenience layer to make response declaration simpler. For example:

```go
// This:
responses.RedirectPermanently()

// Is equal to:
huma.NewResponse(http.StatusRedirectPermanently, http.StatusText(http.StatusRedirectPermanently))
```

Error responses also add a content type and model for you. Overall the goal is to improve readability and easy-of-use, but you can fall back to `huma.NewResponse` for more complex scenarios if needed.

> Since ctx includes a response writer, I'd expect to be able to do something like:
>
> ```
> ctx.response.BadRequest("BadInput")
> ```
> Which looks more similar to the declared response

It's specifically an [`http.ResponseWriter`](https://golang.org/pkg/net/http/#ResponseWriter) so it only supports `Header()`, `WriteHeader`, and `Write` calls just like the standard library. We only add the `WriteModel` and `WriteError` calls for convenience. We could add more convenience and increase the size of the interface, but in general you want to keep the interfaces small. I'm curious to hear if others find this confusing.

README.md
```diff
@@ -25,464 +24,423 @@ Features include:
...
+thing := app.Resource("/things/{thing-id}")
+
+// Next, declare the handler with an input argument.
+thing.Get("get-thing", "Get a single thing",
+	responses.NoContent(),
+).Run(func(ctx huma.Context, input MyInput) {
+	fmt.Printf("Thing ID: %s\n", input.ThingID)
+	fmt.Printf("Query param: %s\n", input.QueryParam)
+	fmt.Printf("Header param: %s\n", input.HeaderParam)
+	fmt.Printf("Body name: %s\n", input.Body.Name)
```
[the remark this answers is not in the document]
danielgtaylor: I agree this is somewhat different from those other frameworks, and for the most part this is on purpose despite it introducing a (hopefully small) learning curve. Parameters are a good example, so let's take the Express example because I love that framework:

```js
app.get('/user/:user', function(req, res, next){
  res.send('user ' + req.user.name);
});
```

What type is `:user`? What sort of validation is on that parameter? Is there any documentation for it? Is it an ID, a name, something else? If you look earlier in the code you see this:

```js
app.param('user', function(req, res, next, id){
  if (req.user = users[id]) {
    next();
  } else {
    next(createError(404, 'failed to find user'));
  }
});
```

So there is some validation actually, and a very non-obvious possibility of getting a 404 outside of the handler itself. Outside of *maybe* really complex static code analysis you have no way of determining this is the case. You couldn't generate documentation off it. It's not self-descriptive or obvious.

Part of Huma's reason for existing is that I'm saying the above doesn't meet my minimum bar for maintainable, self-descriptive service code or a user-friendly client experience unless they also have hand-written up-to-date docs somewhere. In Huma we try to make it much more obvious, and it does use the param in the request URI template because we need to know the position of it:

```go
type UserParam struct {
  User string `path:"user-name" pattern="^[a-z0-9-]+$" doc:"Unique user name"`
}

app := cli.NewRouter("Test Router", "1.0.0")

app.Resource("/users/{user-name}").Get("get-user", "Get a user by name",
  responses.OK().Model(User),
  responses.NotFound(),
).Run(func(ctx huma.Context, input UserParam) {
  u, err := db.GetUser(input.User)
  if err != nil {
    ctx.WriteError(http.StatusNotFound, input.User + " user not found")
    return
  }

  ctx.WriteModel(http.StatusOK, u)
})
```

Now we get response info, param type information & safety, validation through a regex, documentation for OpenAPI (docs, mocks, SDKs, CLIs, etc), and hopefully straightforward code out of the box. My hope is that these things are worth the learning curve coming from other frameworks.

context_test.go
```diff
@@ -0,0 +1,244 @@
...
+		ctx.WriteModel(http.StatusOK, R2{Bar: "blah"})
+	})
+
+	assert.Panics(t, func() {
+		w := httptest.NewRecorder()
+		req, _ := http.NewRequest(http.MethodGet, "/bad-status", nil)
+		app.ServeHTTP(w, req)
+	})
+
+	assert.Panics(t, func() {
```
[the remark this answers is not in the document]
weiser: n/m this. I see we can do such a thing by applying the `Recovery` middleware.

README.md
```diff
@@ -25,464 +24,423 @@ Features include:
...
+	Name string `query:"name"`
+}
+
+func (m *MyInput) Resolve(ctx huma.Context, r *http.Request) {
+	// Get request info you don't normally have access to.
+	m.Host = r.Host
+
+	// Transformations or other data validation
+	m.Name = strings.Title(m.Name)
+}
```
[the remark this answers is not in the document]
danielgtaylor: Yes, that's how you would use it. I'll update the example! :+1:

README.md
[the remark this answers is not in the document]
weiser: > They all deliberately match, so responses\.(.*) ⇨ http.Status$1, e.g. http.StatusRedirectPermanently. 

Something to this effect should be added to the readme

README.md
```diff
@@ -511,155 +469,75 @@ The standard `json` tag is supported and can be used to rename a field and mark
...
-Huma includes a dependency injection system that can be used to pass additional arguments to operation handler functions. You can register global dependencies (ones that do not change from request to request) or contextual dependencies (ones that change with each request).
-
-Global dependencies are created by just setting some value, while contextual dependencies are implemented using a function that returns the value of the form `func (deps..., params...) (headers..., *YourType, error)` where the value you want injected is of `*YourType` and the function arguments can be any previously registered dependency types or one of the hard-coded types:
+## Middleware
 
-- `huma.ConnDependency()` the current `http.Request` connection (returns `net.Conn`)
-- `huma.ContextDependency()` the current `http.Request` context (returns `context.Context`)
-- `huma.GinContextDependency()` the current Gin request context (returns `*gin.Context`)
-- `huma.OperationIDDependency()` the current operation ID (returns `string`)
+Standard Go middleware is supported. It can be attached to the main router/app or to individual resources, but **must** be added _before_ operation handlers are added.
```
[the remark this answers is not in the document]
danielgtaylor: Unfortunately it seems to just be a de facto standard that everyone uses without an official type definition in the standard library. I did find this article from 2013 and it's all still the same in 2020:

- https://justinas.org/writing-http-middleware-in-go
- https://www.alexedwards.net/blog/making-and-using-middleware

I'm not sure what the best way to link to this concept is...

README.md
```diff
@@ -672,119 +550,144 @@ r.Resource("/timeout",
...
 ```go
-r := huma.NewRouter("My API", "1.0.0")
+app := cli.NewRouter("My API", "1.0.0")
 
+create := app.Resource("/foo").Post("create-item", "Create a new item",
+	responses.NoContent(),
+)
 // Limit set to 10 MiB
-r.Resource("/foo", MaxBodyBytes(10 * 1024 * 1024)).Get(...)
+create.MaxBodyBytes(10 * 1024 * 1024)
```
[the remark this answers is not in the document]
danielgtaylor: You get this (serialized to JSON in the response), since we stop reading and may not know the full input size. Length might come from the request's `Content-Length` header or it might be zero if unknown:

```go
ctx.AddError(&ErrorDetail{
  Message:  fmt.Sprintf("Request body too large, limit = %d bytes", ctx.op.maxBodyBytes),
  Location: "body",
  Value:    length,
})
```

From https://github.com/danielgtaylor/huma/pull/11/files#diff-6756fdfda338c13b910a6416f2ac9f54R194

README.md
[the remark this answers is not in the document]
weiser: Maybe just link to those?

context_test.go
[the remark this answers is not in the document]
danielgtaylor: Good thinking. This actually isn't obvious in the test, but we are creating the router **without** the recovery middleware since you might write your own. That makes it easy to assert that it would panic in certain situations and makes writing the tests easier along with preventing circular imports. I'll try to add a comment about that.

In the middleware package there is a test to assert you do in fact get a 5xx, e.g. see:

```go
func TestRecoveryMiddleware(t *testing.T) {
	app, _ := newTestRouter(t)

	app.Resource("/panic").Get("panic", "Panic recovery test",
		responses.NoContent(),
	).Run(func(ctx huma.Context) {
		panic(fmt.Errorf("Some error"))
	})

	w := httptest.NewRecorder()
	req, _ := http.NewRequest(http.MethodGet, "/panic", nil)
	app.ServeHTTP(w, req)
	assert.Equal(t, http.StatusInternalServerError, w.Code)
	assert.Equal(t, "application/problem+json", w.Result().Header.Get("content-type"))
}
```

https://github.com/danielgtaylor/huma/pull/11/files#diff-ddd4a23b5d23539d33f99dcf0289088fR45

danielgtaylor:
> Does any public API break as a result of this refactor?

It definitely breaks. We are still in the v0 phase so we can technically break, but a doc describing how to migrate would be great. After I get general approval here I want to migrate on of our services and write the doc based on that work.

weiser:
If there are breaks, I've got a project that I can use as a guinea pig to contribute a migration strategy (although the migration strategy would be scoped to only changes I had to make to get my project to work)/

context.go
```diff
@@ -0,0 +1,231 @@
...
+
+			for _, h := range allowed {
+				if strings.ToLower(name) == strings.ToLower(h) {
+					found = true
+					break
+				}
+			}
+
+			if !found {
+				panic(fmt.Errorf("Header %s is not declared for %s %s (allowed: %s)", name, c.r.Method, c.r.URL.Path, allowed))
```
marcind, line 88: Missing context that this header was not allowed for a particular status code.

context.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (c *hcontext) AddError(err error) {
+	c.errors = append(c.errors, err)
+}
+
+func (c *hcontext) HasError() bool {
+	return len(c.errors) > 0
+}
+
+func (c *hcontext) WriteHeader(status int) {
```
marcind, line 65: Never understood why methods with this purpose are called "WriteHeader" in golang-world. You're not writing a "header" (as in a request or response header), you are writing a status code (or [status line](https://tools.ietf.org/html/rfc7230#section-3.1.2)). I find the name confusing.

marcind: But I guess this implements `http.ResponseWriter` so 🤷

context.go
```diff
@@ -0,0 +1,231 @@
...
+func (c *hcontext) Write(data []byte) (int, error) {
+	if c.closed {
+		panic(fmt.Errorf("Trying to write to response after WriteModel or WriteError for %s %s", c.r.Method, c.r.URL.Path))
+	}
+
+	return c.ResponseWriter.Write(data)
+}
+
+func (c *hcontext) WriteError(status int, message string, errors ...error) {
+	details := []*ErrorDetail{}
```
marcind, line 105: check that the context is `closed` just like in https://github.com/danielgtaylor/huma/pull/11/files#diff-05a3aa0ab9e5687ff167e0ef3ae4c5a6R97

context.go
```diff
@@ -0,0 +1,231 @@
...
+		ct = "application/problem+json"
+	case "application/yaml", "application/x-yaml":
+		ct = "application/problem+yaml"
+	}
+
+	c.writeModel(ct, status, model)
+}
+
+func (c *hcontext) WriteModel(status int, model interface{}) {
+	// Get the negotiated content type the client wants and we are willing to
```
marcind, line 138: check that the context is closed just like in https://github.com/danielgtaylor/huma/pull/11/files#diff-05a3aa0ab9e5687ff167e0ef3ae4c5a6R97

error.go
```diff
@@ -1,6 +1,30 @@
...
+
+// ErrorDetailer returns error details for responses & debugging.
+type ErrorDetailer interface {
+	ErrorDetail() *ErrorDetail
+}
+
+// ErrorDetail provides details about a specific error.
+type ErrorDetail struct {
+	Message  string      `json:"message,omitempty"`
+	Location string      `json:"location,omitempty"`
```
marcind, line 13: What's the purpose of Location? Specifically, is it a source code location? Because an end-user won't be able to do anything with that information (other than providing it in a support ticket, in which case some sort of error code would probably be better anyway).

examples/echo/echo.go
```diff
@@ -0,0 +1,77 @@
...
+			responses.BadRequest(),
+		).
+		//WithDeadline(30 * time.Second).
+		Run(func(ctx huma.Context, input EchoRequest) {
+			v := input.Word
+			if input.Greet {
+				v = "Hello, " + v
+			}
+
+			ctx.Header().Set("Etag", `W/"foo"`)
```
marcind, line 69: Canonically it's

```suggestion
			ctx.Header().Set("ETag", `W/"foo"`)
```

humatest/humatest.go
```diff
@@ -16,22 +16,23 @@ import (
 // NewRouter creates a new test router. It includes a logger attached to the
 // test so if it fails you will see additional output. There is no recovery
 // middleware so panics will get caught by the test runner.
```
marcind, line 18: I actually find that very annoying, because I _think_ that if a single test fails you'll get log spew from all the tests (making debugging harder). This would be even worse for table-drive tests where maybe just one case is failing. I would like to be able to at least have the option to not attach a logger.

middleware/middleware.go
```diff
@@ -0,0 +1,54 @@
...
+	// Note: logger goes before recovery so that recovery can use it. We don't
+	// expect the logger to cause panics.
+	return chi.Chain(
+		OpenTracing,
+		Logger,
+		Recovery(func(ctx context.Context, err error, request string) {
+			log := GetLogger(ctx)
+			log = log.With(zap.Error(err))
+			log.With(
+				zap.String("http.request", string(request)),
```
marcind, line 34: `request` is already a string, no?

middleware/logger.go
```diff
@@ -0,0 +1,132 @@
...
+
+		chiCtx := chi.RouteContext(r.Context())
+
+		contextLog := l.With(
+			zap.String("http.version", r.Proto),
+			zap.String("http.method", r.Method),
+			// The route pattern isn't filled out until *after* the handler runs...
+			// zap.String("http.template", chiCtx.RoutePattern()),
+			zap.String("http.url", r.URL.String()),
+			zap.String("network.client.ip", r.RemoteAddr),
```
marcind, line 87: For integration with DataDog APM, it's nice to attach `dd.trace_id` and `dd.span_id` with values from the current trace context so that APM traces can link to log messages.

middleware/minimal.go
```diff
@@ -0,0 +1,45 @@
...
+	if w.status == http.StatusNoContent {
+		return 0, nil
+	}
+
+	return w.ResponseWriter.Write(data)
+}
+
+func (w *minimalWriter) WriteHeader(statusCode int) {
+	if statusCode >= 200 && statusCode < 300 {
+		statusCode = http.StatusNoContent
```
marcind, line 26: https://tools.ietf.org/html/rfc7240#section-4.2 has an example case where a 201 Created is an appropriate status code.

middleware/opentracing.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"net/http"
+
+	"github.com/go-chi/chi"
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+)
+
+// OpenTracing provides a middleware for cross-service tracing support.
+func OpenTracing(next http.Handler) http.Handler {
```
marcind, line 12: Would be nice to see an example of how this integrates with the datadog tracer API.

context.go
danielgtaylor: Yeah this is purely to match up with the standard library `http.ResponseWriter` interface so we don't really control it :cry:

error.go
danielgtaylor: I've added some additional documentation. This is actually the location in the input as a JSONPath type link and can refer to request parameters (path/query/header) or the request body. For example:

```json
{
  "status": 400,
  "title": "Bad Request",
  "detail": "Request input did not pass validation",
  "errors": [
    {
      "message": "Could not parse int",
      "location": "path.thing-id",
      "value": "abc"
    },
    {
      "message": "Field is required",
      "location": "body.tags[1].description"
    },
    {
      "message": "Must match pattern ^[a-z0-9-]+$",
      "location": "body.slug",
      "value": "foo_bar"
    }
  ]
}
```

humatest/humatest.go
danielgtaylor: You have a couple of options for this:

1. Call `humatest.NewRouter` for each test in the table, then you get only the single failing log output.
1. Ignore the `humatest` package and just do `huma.New` instead and you get a blank router without any middleware.

middleware/middleware.go
danielgtaylor: Unfortunately it's `[]byte` so we have to cast it.

context.go
marcind: I think it actually ends up doing the check because it will ultimately go call `writeModel` which calls `Write` which does the check, but before that happens it might end up doing a bunch of work. Not a big deal though.

middleware/middleware.go
marcind: Isn't it a `string` here https://github.com/danielgtaylor/huma/pull/11/files/f1f9c8507156c28ef756697193edc96fcbb51cea#diff-ce31218ed9b14d108ecc98733ac0abd2R30 ?

middleware/opentracing.go
danielgtaylor: I think this is all you have to do:

```go
// Start a Datadog tracer, optionally providing a set of options,
// returning an opentracing.Tracer which wraps it.
t := opentracer.New(tracer.WithAgentAddr("host:port"))

// Use it with the Opentracing API. The (already started) Datadog tracer
// may be used in parallel with the Opentracing API if desired.
opentracing.SetGlobalTracer(t)
```

https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/opentracer/example_test.go#L15-L23

middleware/middleware.go
danielgtaylor: Oh, duh, it's already cast in the `Recovery` middleware itself. This was a copy/paste error. Will fix.

README.md
```diff
@@ -25,464 +24,431 @@ Features include:
...
+// Create and start a new router by hand:
+router := huma.New("API Name", "1.0.0")
+router.Middleware(middleware.DefaultChain)
+router.Listen("127.0.0.1:8888")
 ```
 
 ## Resources
 
-Huma APIs are composed of resources and sub-resources attached to a router. A resource refers to a unique URI on which operations can be performed. Huma resources can have dependencies, security requirements, parameters, response headers, and responses attached to them which are all applied to every operation and sub-resource.
+Huma APIs are composed of resources and sub-resources attached to a router. A resource refers to a unique URI on which operations can be performed. Huma resources can have middleware attached to them, which run before operation handlers.
```
lgarrett-isp, line 138: > which run before operation handlers.
Is my understanding correct that the middleware executes to completion, then the operation handler is executed? 

No change requested, just looking to understand: I am mentally contrasting it with a gRPC interceptor where the middleware _calls_ the handler so it can also take action on the handler response.

benchmark/go.mod
```diff
@@ -0,0 +1,12 @@
+module github.com/danielgtaylor/huma/benchmark
+
+go 1.14
+
+replace github.com/danielgtaylor/huma => ../
```
lgarrett-isp, line 5:
```suggestion
```

README.md
danielgtaylor: It's the same concept. Each middleware takes a `next` handler and calls `next.ServeHTTP(w, r)` on it, and at the end of the chain `next` is just the handler function itself.

middleware/middleware.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/go-chi/chi"
+	"go.uber.org/zap"
+)
+
+// Middlewarer lets you add middlewares
```
ktracer-isp, line 11: This feels like an odd naming convention.

Also, is there a way to specify the order of middleware layers or are we assuming they're idempotent?

middleware/recovery.go
```diff
@@ -0,0 +1,129 @@
...
+
+	"github.com/danielgtaylor/huma"
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+)
+
+// MaxLogBodyBytes logs at most this many bytes of any request body during a
+// panic when using the recovery middleware. Defaults to 10KiB.
+var MaxLogBodyBytes int64 = 10 * 1024
+
```
ktracer-isp, line 23: This feels like the kind of value we may want to make configurable per request, but right now it's a one-time set-and-forget.

middleware/recovery.go
```diff
@@ -0,0 +1,129 @@
...
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			var buf *bytes.Buffer
+
+			// Reset the body so other middleware/handlers can use it.
+			if r.Body != nil {
+				// Get a buffer that the body will be read into.
+				buf = bufPool.Get().(*bytes.Buffer)
+				defer bufPool.Put(buf)
+
+				r.Body = newBufferedReadCloser(r.Body, buf, MaxLogBodyBytes)
```
ktracer-isp, line 87: If this is a panic, we should probably ensure we're logging the whole request even if it's bigger than 10k bytes.

middleware/recovery.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+
+			// Recovering comes *after* the above so the buffer is not returned to
+			// the pool until after we print out its contents.
+			defer func() {
```
ktracer-isp, line 92: Anonymous goroutines are hard to test independently.

negotiation/negotiation.go
```diff
@@ -0,0 +1,50 @@
+package negotiation
+
+import (
+	"strconv"
+	"strings"
+)
+
+// SelectQValue selects and returns the best value from the allowed set
+// given a header with optional quality values, as you would get for an
+// Accept or Accept-Encoding header. The *first* item in allowed is preferred
```
ktracer-isp, line 10: Do we expect people to send ordered lists of preference, or weighted lists? This code feels like it's trying to split the difference and would benefit from establishing one call pattern or the other.

operation.go
```diff
@@ -0,0 +1,247 @@
...
+// form: `func (ctx huma.Context)` or `func (ctx huma.Context, input)` where
+// input is your input struct describing the input parameters and/or body.
+func (o *Operation) Run(handler interface{}) {
+	if reflect.ValueOf(handler).Kind() != reflect.Func {
+		panic(fmt.Errorf("Handler must be a function taking a huma.Context and optionally a user-defined input struct, but got: %s for %s %s", handler, o.method, o.resource.path))
+	}
+
+	var register func(string, http.HandlerFunc)
+
+	switch o.method {
```
ktracer-isp, line 149: Do we need a default that panics or otherwise indicates an unexpected verb?

context.go
```diff
@@ -0,0 +1,239 @@
...
+	// optionally a set of error details set with `AddError` and/or passed in.
+	WriteError(status int, message string, errors ...error)
+
+	// WriteModel writes out an HTTP status code and marshalled model based on
+	// content negotiation (e.g. JSON or CBOR). This must match the registered
+	// response status code & type.
+	WriteModel(status int, model interface{})
+}
+
+type hcontext struct {
```
ktracer-isp, line 48: This isn't a very informative type name; what is meant by `hcontext`?

resolver.go
```diff
@@ -0,0 +1,391 @@
...
+			pv = v
+		}
+
+		if name, ok := f.Tag.Lookup("path"); ok {
+			pname = name
+			location = "path"
+			if v := chi.URLParam(req, name); v != "" {
+				pv = v
+			}
+		}
```
ktracer-isp, line 249: You have this logic repeated three times; is this a switch in disguise?

resource.go
```diff
@@ -2,194 +2,104 @@ package huma
...
 }
 
-// Patch creates an HTTP PATCH operation on the resource.
-func (r *Resource) Patch(docs string, handler interface{}) {
-	r.operation(http.MethodPatch, docs, handler)
-}
+// SubResource creates a new resource attached to this resource. The passed
+// path will be appended to the resource's existing path. The path can
+// include parameters, e.g. `/things/{thing-id}`. Each resource path must
+// be unique.
```
ktracer-isp, line 87: Is there anything enforcing uniqueness, or is this a convention?

responses/responses.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// GatewayTimeout HTTP 504 response with a structured error body (e.g. JSON).
+func GatewayTimeout() huma.Response {
+	return errorResponse(http.StatusGatewayTimeout)
+}
+
+// String HTTP response with the given status code and `text/plain` content
+// type.
+func String(status int) huma.Response {
```
ktracer-isp, line 133: Is there a way to return a JSON-formatted error response? This is something we've had requested.

benchmark/fiber/main.go
```diff
@@ -0,0 +1,48 @@
+package main
```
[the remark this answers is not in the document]
danielgtaylor: This is just a benchmark against another similar library. It's explicitly excluded from tests/coverage and such. Purely for benchmarking.

benchmark/huma/main.go
```diff
@@ -17,34 +18,30 @@ type Item struct {
 	IsOffer bool    `json:"is_offer,omitempty"`
 }
 
+type Input struct {
+	AuthInfo string
+	ID       int `path:"id"`
+}
+
+func (i *Input) Resolve(ctx huma.Context, r *http.Request) {
+	i.AuthInfo = strings.Split(r.Header.Get("Authorization"), " ")[0]
```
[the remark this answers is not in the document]
danielgtaylor: It's described in the benchmark folder's README, and for the most part the format doesn't matter. It's mainly about testing the various parameter parsing code in different libraries.

context.go
```diff
@@ -0,0 +1,239 @@
...
+
+	// Do the appropriate encoding.
+	var encoded []byte
+	var err error
+	if strings.HasPrefix(ct, "application/json") || strings.HasSuffix(ct, "+json") {
+		encoded, err = json.Marshal(model)
+		if err != nil {
+			panic(fmt.Errorf("Unable to marshal JSON: %w", err))
+		}
+	} else if strings.HasPrefix(ct, "application/yaml") || strings.HasPrefix(ct, "application/x-yaml") || strings.HasSuffix(ct, "+yaml") {
```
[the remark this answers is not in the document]
danielgtaylor: [RFC 6839 Section 3.1](https://tools.ietf.org/html/rfc6839#section-3.1) says that:

>[RFC4627] defines the "application/json" media type.  The suffix "+json" MAY be used with any media type whose representation follows that established for "application/json".

The same is true for the `+yaml` suffix. I guess that means `application/json+yaml` might be valid, but in practice nobody uses it and it *should* be parsed as YAML.

docs.go
```diff
@@ -19,10 +18,11 @@ func splitDocs(docs string) (title, desc string) {
...
 
-// rapiDocTemplate is the template used to generate the RapiDoc.  It needs two args to render:
-// 1. the title
-// 2. the path to the openapi.yaml file
-var rapiDocTemplate = `<!doctype html>
+// RapiDocHandler renders documentation using RapiDoc.
+func RapiDocHandler(router *Router) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "text/html")
+		w.Write([]byte(fmt.Sprintf(`<!doctype html>
```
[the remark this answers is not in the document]
danielgtaylor: Can you give an example of how this should look? I'm having trouble understanding this comment :frowning_face:

middleware/logger.go
```diff
@@ -0,0 +1,152 @@
...
+
+// LogTracePrefix is used to prefix OpenTracing trace and span ID key names in
+// emitted log message tag names. Use this to integrate with DataDog and other
+// tracing service providers.
+var LogTracePrefix = "dd."
+
+// NewDefaultLogger returns a new low-level `*zap.Logger` instance. If the
+// current terminal is a TTY, it will try ot use colored output automatically.
+func NewDefaultLogger() (*zap.Logger, error) {
+	if LogLevel != nil {
```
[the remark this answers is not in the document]
danielgtaylor: I can see a potential use-case for this, but I'd like to argue that for the *built-in* logger we keep things simple. You can always bring your own more complex logger. I think most microservices we are designing and working on wouldn't benefit much from a more complex logger setup given their limited scope and small codebase. If that changes, we can bring our own more complicated logger as needed on a case-by-case basis.

The current iteration of the default logger middleware has one global level that you can modify at runtime, so we can still enable e.g. debug logs dynamically if we wanted to with this design.

middleware/middleware.go
danielgtaylor: It's odd but it is idiomatic Go, e.g. see https://golang.org/doc/effective_go.html#interface-names

middleware/recovery.go
danielgtaylor: You can actually set this because it's a public var. You can do `middleware.MaxLogBodyBytes = 10 * 1024 * 1024` to log 10MB before your service starts.

middleware/recovery.go
danielgtaylor: This is probably worth a larger discussion. When designing a web framework you have to balance speed / memory use / concurrent connection capacity / etc. Currently this uses a shared auto-scaling pool of byte buffers that read up to a certain amount into memory for every single incoming request, and that's not cheap. Given the size of most API requests 10KB was chosen as a sane default that should print the whole request most of the time, but prevents large bodies from taking down the service. This is definitely a compromise that attempts to give us at least some extremely useful debugging data at the expense of some speed and higher memory thrashing. Benchmarks show that we can process a couple thousand fewer requests a second because of it (a few percent).

I understand our current use-case isn't extreme high performance, but keep in mind this is meant to be a general-purpose web framework and could see adoption across many teams and companies.

middleware/recovery.go
danielgtaylor: This is just the standard way to do recovery and it is tested in `recovery_test.go`.

context.go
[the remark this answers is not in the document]
ktracer-isp: Gotcha; I wasn't sure if there was the possibility of needing to support multiple return types and the way it was written made it seem like it could be. Good to know it's not!

negotiation/negotiation.go
danielgtaylor: The "Q" value is an HTTP concept. See https://developer.mozilla.org/en-US/docs/Glossary/Quality_values. This code takes the user-supplied weighted list and picks from a developer-supplied ordered list to get a match for content negotiation. The developer-supplied ordered list is only ordered to break a tie should multiple user-supplied items with the same weight match. I don't think using weights for the developer side makes sense in this scenario and is more complicated than it needs to be.

docs.go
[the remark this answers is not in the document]
ktracer-isp: Ah, sorry. There are blocks of HTML Scattered through the tests that look mostly similiar, and I thought it would improve readability to pull those HTML snippets into a fixture file of some kind. Does that make more sense?

context.go
danielgtaylor: The name `context` is taken by a standard library package, so we append the name of this package. I guess that means it stands for "Huma context". It was just a way to prevent shadowing and is an internal implementation detail of the public `huma.Context` interface.

middleware/logger.go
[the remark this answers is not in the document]
ktracer-isp: That all sounds reasonable then; the one feature request I'd like to add to the Huma backlog, then, is a way to dynamically change logging level without having to restart the code. That's future work, though, not this refactor.

resolver.go
danielgtaylor: You have to call `Lookup(...)` with a different tag name each time, so I wasn't able to make this a `switch`. Maybe I'm missing something though?

middleware/recovery.go
ktracer-isp: Ah, gotcha. I'm still not used to Go's scoping algorithm. "Capitalize to make public" really messes with me. xvx

resource.go
danielgtaylor: The underlying router will enforce it and panic when you try to register the same one twice. If this is too much of a burden, we could cache the sub-routers for re-use, but I'd like to tackle that as a future change if we run into it.

responses/responses.go
danielgtaylor: Yep, any time you use `ctx.AddError` or `ctx.WriteError` you will get a nicely formatted [RFC 7807](https://tools.ietf.org/html/rfc7807) JSON error returned. If you use the built-in JSON Schema-based validation you even get enhanced errors with the specific JSON Path location and value along with why the validation failed at that location, e.g. `body.foo.bar` and `Unable to parse boolean: abc`.

middleware/recovery.go
ktracer-isp: If the impact of changing this value is that big, this feels like the kind of switch that comes with a warning label.

I don't know if there's anything in the README about it, but a comment about the impact that changing this value could have on system performance might be worth including. If it's already there, apologies on my having missed it.

middleware/logger.go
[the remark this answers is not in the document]
danielgtaylor: I'm pretty sure you can do that with this PR by doing `middleware.LogLevel = zap.DebugLevel` for example. It should all just work for the default logging middleware.

middleware/recovery.go
danielgtaylor: Yeah it's definitely something to get used to :smile:

docs.go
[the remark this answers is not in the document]
danielgtaylor: I think each of these blocks is unique and comes from different documentation project's READMEs/demos. I'll take a look around for anything that might be duplicated but for the most part keeping these separate makes them easier to update (just a copy/paste from upstream if/when needed).

middleware/recovery.go
ktracer-isp: I accept that a `defer func()` is the standard way to do recovery. I'm talking more broadly about the fact that nested anonymous functions can't be tested in isolation of their surrounding code. That to me is a code smell, especially when the fix is as simple as "declare the function elsewhere in the code and reference it by name." I accept that sometimes this is going to happen in places where we really do intend these components really _shouldn't_ be decomposed, but I'd prefer it if in our codebase those are the exception and called out with comments, rather than the rule.

negotiation/negotiation.go
ktracer-isp: That's fair. My original comment was more that this feels like an overoptimization for a very rare use case. That said, it's not bad code, and given the potential reach of the codebase, I don't think it should be removed. This section just feels... I don't know, weirdly fiddly when compared to its surrounding complexity. I don't have any specific action items here, either, so this can be resolved if it's just "this is how it needs to operate to be compliant with the spec."

resolver.go
ktracer-isp: I was apparently quite scattered when I wrote `switch` because looking at it again, I'm not sure what I meant.

I _think_ what I was trying to ask was whether we wanted a helper function to perform this function on a repeated basis, but since one of the three blocks appears likely to change in the future, on review, I'm not sure there's nothing here that needs to be done.

PR #26, opened 2022-02-23
fix: copy router into sub-resources

danielgtaylor:
Tiny bug fix to make sure sub-resources get all fields copied. This gets used later when declaring operations on the sub-resources.


lgarrett-isp (approved):
Good catch!

PR #29, opened 2022-03-14
feat: enable applications to redact headers from recovery handler

lgarrett-isp:
Super simple header redaction for the recovery handler before it parses the request to a string and sends it to the application.

### DIFF
--- a/middleware/recovery.go
+++ b/middleware/recovery.go
@@ -78,6 +78,13 @@ func (r *bufferedReadCloser) Close() error {
 	return r.reader.Close()
 }
 
+// RemovedHeaders defines a list of HTTP headers that will be redacted from the
+// request in the Recovery handler--if any logging or other output occurs, these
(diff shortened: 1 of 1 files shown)


middleware/recovery.go
```diff
@@ -78,6 +78,13 @@ func (r *bufferedReadCloser) Close() error {
 	return r.reader.Close()
 }
 
+// RemovedHeaders defines a list of HTTP headers that will be redacted from the
+// request in the Recovery handler--if any logging or other output occurs, these
+// headings will have value '<redacted>'.
+var RemovedHeaders []string
```
danielgtaylor, line 84: Should we default this to `[]string{"Authorization"}` as that header is always going to contain sensitive information?

lgarrett-isp: Agreed! Fixed!

danielgtaylor (approved):

PR #34, opened 2022-03-23
fix: redact headers after handler runs; fix typo

danielgtaylor:
This PR waits to redact sensitive headers until just before they get dumped by the panic recovery middleware logger. Also fixes a minor typo I found the other day.


lgarrett-isp (approved):
lgtm!

PR #36, opened 2022-04-05
Feat: Added WriteContent to huma context in order to support reading from streams directly into http response

deo986:
This is useful for large data requests, especially ones that need to be chunked into multiple range requests like direct video streaming.  Since `http.ServeContent` already does exactly this and meticulously follows the associated RFCs, I decided to write this as a wrapper on top of that function.


context.go
```diff
@@ -373,3 +386,21 @@ func selectContentType(r *http.Request) string {
...
+	if c.closed {
+		panic(fmt.Errorf("Trying to write to response after WriteModel, WriteError, or WriteContent for %s %s", c.r.Method, c.r.URL.Path))
+	}
+
+	// name is left blank, this is used by ServeContent to automatically
+	// determine Content-Type.  Huma has opted to have handlers set Content-Type
+	// explicitly rather than introduce this into the method signature as name
+	// is not applicable to every ReadSeeker.  Leaving this blank or setting the
+	// Content-Type on the request disables this functionality anyway.
+	http.ServeContent(c.ResponseWriter, c.r, "", c.modTime, content)
```
danielgtaylor, line 400: I think passing a name for content-type analysis here would be reasonable. Huma handlers must **declare** the `Content-Type` header if they want to set it, but `WriteContent` could be the thing that actually sets it. Then you don't duplicate all that mimetype logic you probably don't care a ton about. Then you can still manually override if you want but the default is to just handle it for you.

Your `responses.ServeContent` already pre-declare the `Content-Type` header, so this should work great together.

The signature could be `WriteContent(name string, content io.ReadSeeker, lastModified time.Time)` or something along those lines.

deo986: I guess both of these comments speak to me not finding the `ServeContent` function signature to be very intuitive.  It has a lot of code paths that kick only when a particular field is the default value which to me seems not very user friendly at all.  I was trying to make it as simplified as possible and provide only 1 way to interact with the function.  Point taken that this might end up forcing people to reimplement the same mimetype logic that `ServeContent` was already doing, so I can add it back and provide comments, but I just wish there was a clearer way to handle this.

danielgtaylor (approved):
LGTM :+1:

PR #41, opened 2022-04-12
feat: return more specific error codes

danielgtaylor:
This PR changes the status code behavior for errors to attempt to return more specific status codes when possible while still striving for exhaustive error responses. As can be seen in the README, the new flow is:

```mermaid
flowchart TD
	Request[Request has errors?] -->|yes| Panic
	Request -->|no| Continue[Continue to handler]
	Panic[Panic?] -->|yes| 500
	Panic -->|no| RequestBody[Request body too large?]
	RequestBody -->|yes| 413
	RequestBody -->|no| RequestTimeout[Request took too long to read?]
	RequestTimeout -->|yes| 408
	RequestTimeout -->|no| ParseFailure[Cannot parse input?]
	ParseFailure -->|yes| 400
	ParseFailure -->|no| ValidationFailure[Validation failed?]
	ValidationFailure -->|yes| 422
	ValidationFailure -->|no| 400
```

The change is fairly small & unobtrusive. Several tests have been updated to reflect the changes. It should not break clients, however your API will have new documented responses when upgrading Huma.

Note: the above flowchart deliberately excludes stuff like `405 Method Not Allowed` because that returns at the low-level router before any of the above executes.

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -276,7 +276,7 @@ func (o *Operation) Run(handler interface{}) {
 			docsPrefix:            o.resource.router.docsPrefix,
 			urlPrefix:             o.resource.router.urlPrefix,
 			disableSchemaProperty: o.resource.router.disableSchemaProperty,
-			errorCodeHint:         http.StatusBadRequest,
+			errorCode:             http.StatusBadRequest,
 		}
 
 		// If there is no input struct (just a context), then the call is simple.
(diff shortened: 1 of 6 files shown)


operation.go
```diff
@@ -296,9 +307,14 @@ func (o *Operation) Run(handler interface{}) {
...
 		setFields(ctx, ctx.r, input, inputType)
+		if !ctx.HasError() {
+			// No errors yet, so any errors that come after should be treated as a
+			// semantic rather than structural error.
+			ctx.errorCodeHint = http.StatusUnprocessableEntity
+		}
 		resolveFields(ctx, "", input)
 		if ctx.HasError() {
-			ctx.WriteError(http.StatusBadRequest, "Error while parsing input parameters")
+			ctx.WriteError(ctx.errorCodeHint, "Error while processing input parameters")
```
aorban-isp, line 317: nit: Should this be `ctx.errorCode` rather than `ctx.errorCodeHint`?  It's directly written at the response, not hinting at it.

aorban-isp (approved):

PR #44, opened 2022-04-22
Huma now accepts naked boolean queryparams

deo986:
Today in order to use a boolean with queryparams it needs to be set like this: `?param=true`, now you can set it with `?param` or `?param`

### DIFF
--- a/resolver.go
+++ b/resolver.go
@@ -275,7 +275,7 @@ func setFields(ctx *hcontext, req *http.Request, input reflect.Value, t reflect.
 			location = locationQuery
 			if v := req.URL.Query().Get(name); v != "" {
 				pv = v
-			} else if req.URL.Query().Has(name) {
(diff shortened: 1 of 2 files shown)


resolver.go
```diff
@@ -275,6 +275,9 @@ func setFields(ctx *hcontext, req *http.Request, input reflect.Value, t reflect.
 			location = locationQuery
 			if v := req.URL.Query().Get(name); v != "" {
 				pv = v
+			} else if req.URL.Query().Has(name) {
+				// name has no associated value, but exists in the map of QueryParams.  This is a boolean value
+				pv = "true"
```
danielgtaylor, line 280: I think we also need to check that the type is a boolean here, otherwise for your example tests below I think going `GET /?s=&b` will actually set the string to `"true"` rather than `""`. It's an edge case for sure, but probably one we want to avoid if possible.

danielgtaylor (approved):
LGTM :+1:

PR #50, opened 2022-05-09
feat: allow hidden/undocumented resources

sslotnick-isp:
Allows for creating hidden resource routes that don't appear in the OpenAPI spec or documentation. This is useful for creating endpoints like LB health checks, readiness checks, etc which can use the same server mux but should not appear as part of the API.


danielgtaylor (approved):
LGTM, thanks! :+1:

Issue #20, opened 2021-02-24
How to add a catch-all route and static asset handling to serve html + assets


MichaelBurgess:
As this is built on top of chi I feel/hope this should be possible, but I'm not seeing any APIs in huma to do this?

Forgive me if there's an obvious solution but I'm still getting to grips with this framework.

MichaelBurgess:
if I could get to the chi router then perhaps I could do this? https://github.com/go-chi/chi/blob/master/_examples/fileserver/main.go

MichaelBurgess:
My ideal setup is really this:

- `/api/<known set up API routes>` have Open API documented routes
- `/api/<anything not declared above>` - 404
- `/assets` - serve up CSS / JS / images etc
- `/*` - any other request, serve index.html (which in turn will reference CSS / JS / etc assets

Is this possible with huma? Looks like I'm only really needing this for the known API routes to get the Open API docs, but need to break out of the model for everything else?

Fishwaldo:
Also looking for this to bundle a simple SPA Admin page. 

As I can see it - Best way to do this right now, is write our own New function to return a Router?

danielgtaylor:
With Huma v2 and the bring-your-own-router approach, this is now possible by just adding whatever route you need to the router itself, including using e.g. https://pkg.go.dev/net/http#FileServer

PR #56, opened 2022-07-26
Add GetMux function to Resource to get the Chi Mux

Fishwaldo:
This semi addresses issue #20 in that it gives us a Getter to get the chi mux so we can add handlers for methods other than Huma Resources.


danielgtaylor:
Closing this in favor of #104. You cannot get the mux, but you can pass one in if you want to have access to it.

PR #61, opened 2022-08-19
add AddAllowedHeaders function to whitelist common headers

Fishwaldo:
Saving having to add Headers method to all the responses. I have both a Auth Framework as well as Tracing middleware thats adding headers to every request.


danielgtaylor (approved):
Thanks! :+1:

PR #67, opened 2022-09-13
fix: panic on put without body

danielgtaylor:
This adds an additional `nil` check to prevent a panic when registering or handling a request for a `PUT` operation without a request body. Also adds a test to exercise the functionality and confirm the auto-patch feature skips such operations.


jmasonISP (approved):
:shipit:

PR #70, opened 2022-09-21
fix: auto-mark patch deprecated

iwong-isp:
This will mark auto-generated `PATCH` operations to also be flagged as `deprecated` if the corresponding `PUT` operation is flagged as so.


danielgtaylor (approved):
LGTM :+1:

Issue #68, opened 2022-09-14
Stable OpenAPI property ordering


danielgtaylor:
Currently Huma generates OpenAPI properties in a random order. A stable order would be preferred as it makes things like diffs much cleaner.

OpenAPI is generated via `gabs` but it doesn't seem to have ordering/sort options available, so another approach will need to be used. Maps in Go always have their keys sorted lexicographically when marshaled into JSON, so that could be used.

danielgtaylor:
Ordering is now handled by the YAML serializer in v2.

PR #71, opened 2022-09-30
fix: stable OpenAPI property and param ordering

iwong-isp:
Addresses two issues with stable/consistent OpenAPI generation:

- Properties are generated in random order. https://github.com/danielgtaylor/huma/issues/68
- Operation parameters may be added in random order. 
   - This is particularly problematic in tools like [OpenAPIGenerator](https://github.com/OpenAPITools/openapi-generator) because parameter ordering matters for SDK generation.

Properties will now output in lexicographical order for consistency. Operation parameters will honor the order they're defined in the struct. 

A new example called `bookstore` is also added to demo a slightly more complex OpenAPI configuration where the schema ordering matters.

### DIFF
--- a/router.go
+++ b/router.go
@@ -348,7 +348,6 @@ func (r *Router) setupDocs() {
 	if !r.mux.Match(chi.NewRouteContext(), http.MethodGet, r.OpenAPIPath()) {
 		r.mux.Get(r.OpenAPIPath(), func(w http.ResponseWriter, req *http.Request) {
 			w.Header().Set("Content-Type", "application/vnd.oai.openapi+json")
-
 			w.Write(spec.Bytes())
 		})
 	}
--- a/router.go
+++ b/router.go
@@ -348,7 +348,10 @@ func (r *Router) setupDocs() {
 	if !r.mux.Match(chi.NewRouteContext(), http.MethodGet, r.OpenAPIPath()) {
 		r.mux.Get(r.OpenAPIPath(), func(w http.ResponseWriter, req *http.Request) {
 			w.Header().Set("Content-Type", "application/vnd.oai.openapi+json")
-			w.Write(spec.Bytes())
+
+			// Convert to JSON to take advantage of keys being sorted lexicographically
+			sBytes, _ := json.Marshal(spec)
+			w.Write(sBytes)
 		})
 	}
 
--- a/router.go
+++ b/router.go
@@ -349,9 +349,7 @@ func (r *Router) setupDocs() {
 		r.mux.Get(r.OpenAPIPath(), func(w http.ResponseWriter, req *http.Request) {
 			w.Header().Set("Content-Type", "application/vnd.oai.openapi+json")
 
-			// Convert to JSON to take advantage of keys being sorted lexicographically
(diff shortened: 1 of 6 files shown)


router.go
```diff
@@ -348,7 +348,10 @@ func (r *Router) setupDocs() {
 	if !r.mux.Match(chi.NewRouteContext(), http.MethodGet, r.OpenAPIPath()) {
 		r.mux.Get(r.OpenAPIPath(), func(w http.ResponseWriter, req *http.Request) {
 			w.Header().Set("Content-Type", "application/vnd.oai.openapi+json")
-			w.Write(spec.Bytes())
+
+			// Convert to JSON to take advantage of keys being sorted lexicographically
+			sBytes, _ := json.Marshal(spec)
```
danielgtaylor, line 353: I think `spec.Bytes()` is [supposed to convert to JSON](https://pkg.go.dev/github.com/Jeffail/gabs/v2#Container.Bytes) - curious if you found some issue with that?

router_test.go
```diff
@@ -605,3 +605,53 @@ func TestRequestContentTypes(t *testing.T) {
...
+		Name string `json:"name"`
+	}
+
+	type opschema struct {
+		Parameters []parameters `json:"parameters"`
+	}
+
+	var p opschema
+	json.Unmarshal(openapi, &p)
+	assert.Equal(t, p.Parameters, []parameters{{Name: "menu-category"}, {Name: "item-id"}})
```
danielgtaylor, line 656: non-blocking: there's a ['JSONEq'](https://pkg.go.dev/github.com/stretchr/testify/assert#JSONEq) that could be useful here. Something like:

```go
assert.JSONEq(t, `[{"name": "menu-category", ...}]`, openapi)`
```

danielgtaylor (approved):
Awesome work! :+1: Ordering by Go struct field definition is a good way to quickly fix this bug. I took a quick peek at the router, resource, operation, request, response and I think this PR fixes the only things preventing stable OpenAPI generation!

Definitely not a blocker, but one improvement I can think of for the future might be to order them based on their appearance in the URL template string, regardless of struct order. That way any inconsistency in the Go code ordering would not impact the generated parameter order given two operations with the same URL template. That's more difficult/complex to implement though and I think the current implementation is good.

router.go
iwong-isp: Took another look and you're right in that it coverts to JSON before outputting the bytes. This was likely from me not understanding how the openapi generation worked initially. Reverted in [44bed11](https://github.com/danielgtaylor/huma/pull/71/commits/44bed11fb0b20f5d4dd281de41f3328aaa828db3).

router_test.go
iwong-isp: I considered using that, but unfortunately, the output of the `gabs.Search` ends at the operation level.  This results in `openapi` containing the full spec for the `get` operation (ex. `responses`, additional properties for a `parameter`). `JSONEq` only returns true if the entire JSON payload matches (ignoring order).  Given we're only interested in using the `name` to check order was correctly preserved, I decided with a temp type to only parse the necessary field.

danielgtaylor (approved):
LGTM :+1:

Issue #75, opened 2022-11-22
Recursion in Schemas


mtiller:
I didn't see any notes in the docs about how to handle recursion in schemas.  JSON Schema itself can handle this by using schema references.  But `huma` doesn't seem to be able to recognize the recursion because if I have a model with recursion (with the recursive field marked as `omitempty`), `huma` crashes when accessing an endpoint associated with that model.

danielgtaylor:
Huma doesn't currently support recursive schemas. I'd be happy to review a PR that adds it :+1:

mtiller:
I can see two approaches.  One would be to try and deal directly with recursion.  I can see two possibilities for dealing with recursion.  The first would be to recognize the recursion and use a `$ref` in that case.  But that could be tricky to identify (A includes B includes A kinds of cases).  Another possibility for dealing with recursion would be to use `$ref` for all nested structs.

The other approach I could imagine (and not mutually exclusive with directly dealing with recursion) would be to have a special interface like `Schemer` that could return a `Schema` for a given type.  In this way, if a given type is perhaps complex (_i.e.,_ includes recursion) or the built-in schema generation isn't optimal an explicit (vs. auto-generated) schema could be provided by the type itself.

I took a glance at the code but I just wanted to check a) if you had any preferences with regards to these options and b) whether any of these ideas have already been implemented and I just missed them.

mt35-rs:
Please note, #86 includes an important fix related to this functionality.  Sorry for overlooking that.  The original fix prevents a crash, but it was "incomplete".  The new PR addresses two key issues...the fact that some schemas couldn't be fetched via `/schemas/...` and that `/openapi.json` was missing some scheme definitions.  The fix is pretty simple and the quality is much better as a result.

Issue #74, opened 2022-11-22
Issue with Generics


mtiller:
If I have a model that is generic, then there seems to be an issue resolving the schemas.  For example, if my model is of type `Container[Thing]` and I create my route with:

```go
registry.Get("registry", "Get the registry",
		responses.OK().Model(Container[Thing]{})).Run(func(ctx huma.Context) {

		// The `WriteModel` convenience method handles content negotiation and
		// serializaing the response for you.
		entity := Container[Thing]{}
		entity.Properties = Thing{Name: "Test1"}

		ctx.WriteModel(http.StatusOK, entity)
	})
```

Then my response yields a `Link` header of `</schemas/Container[main.Thing].json>; rel="describedby"`.  By itself, I don't think this is an issue.  But if I try to access the schema that that URL, I get `404 Not Found`.  Presumably this has something to do with the escaping of the characters in the URL or the pattern matching in the router.  Not really sure, but it doesn't find it.

My current solution is to just alias the type in Go, _e.g.,_

```go
type ContainerThing Container[Thing]
```

This seems to resolve the issue.  But I'm not sure I'll have 100% control over such things in nested types.

mt35-rs:
It appears that the issue actually is related to `.` characters in the type name (my example was contrived).  I'm not 100% sure of this, but I'm working on using a regexp in the route for schemas so that the `schema-id` parameter can match examples like `Container[Thing]` (which I think I've succeeded at) as well as `Container[pkg.Thing]` which is causing an issue (I think because the route pattern has a `.` following the route parameter and so it isn't quite getting the right idea about what the prefix is when processing the route tree in `chi`.

In any case, there appears to be another issue (I'll hopefully verify this once I get through this issue).  If I have a type `Container` from a package `pkg` and it is generic and the type parameter is of type `Thing`, the type name constructed by `huma` appears to be `Container[pkg.Thing]` when it should be `pkg.Container[Thing]`.  As I said, I'll have to confirm this and, if true, I'll open a separate issue for this.

mt35-rs:
So, I went down a rabbit hole learning about `chi` routing and I came to the conclusion that the simplest way to deal with this was to simply use a wildcard, `*`, for the schema name and then parse everything in the handler.  Along the way I revamped the error handler as well.  All this is addressed in #81.  I've tested it and it works with all the special characters I've encountered so far.

mtiller:
With #81 merged, I'm closing this.

PR #77, opened 2022-11-30
feat: avoids infinite recursion

mt35-rs:
I'm submitting this pull request in response to issue #75.  

This keeps track (during recursion) of what (`struct`) types have already been defined.  If it encounters an already defined schema, it simply replaces it with an `$ref` rather than elaborating the schema again (and again, and again).

This isn't perfect.  It makes some assumptions about the schema name.

Note, even though the `$ref` I added was something like `./schemas/Foo`, when it gets generated by the schemas endpoint it is `./schemas/Foo.json` which is not good.  Specifically, I'm thinking about cases where we might want to put in a `$ref` that is something like `#/definitions/Foo`.  In that case, no `.json` is required (or desirable).  So you might want to consider revisiting the logic that appends to the `$ref`s.

The other problem here is one that I mentioned in #74 which is that if any of these recursive schemas are generic, then you get a `$ref` that looks something like `./schemas/Struct[parameter].json`.  The `/schemas` endpoint cannot handles these.  So while the schema generation part now works (no more crashes due to infinite recursion), it will almost certainly run into trouble if any client were to actually follow these `$ref`s.

Please let me know if any others concerns you have with this PR.

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -94,8 +94,10 @@ func (c *oaComponents) addSchema(t reflect.Type, mode schema.Mode, hint string, 
 			},
 		}
 	} else {
+		// TODO: See if this type has a predefined schema associated with it
+		// via a predefined map, a callback or an interface.
 		var err error
-		if s, err = schema.GenerateWithMode(t, mode, nil); err != nil {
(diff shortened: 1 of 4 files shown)


danielgtaylor (approved):
I think this looks reasonable to me, thank you!

Issue #79, opened 2022-11-30
Errors have bogus schema


mt35-rs:
The 404 handler generates an RFC 7807 `application/problem+json`, but it doesn't point to a valid schema.  Here is a sample 404 response:

```json
{"$schema":"http://localhost:8888/ErrorModel.json","detail":"Cannot find /ErrorModel.json","status":404,"title":"Not Found"}
```

The `$schema` field _should be_ `http://localhost:8888/schemas/ErrorModel.json`.  Note the presence of `/schemas` in the path.

mt35-rs:
This should be addressed by #80.

PR #80, opened 2022-11-30
fix: this uses proper schema path for errors

mt35-rs:
This propagates various router properties (where possible) and then uses those properties (specifically `schemaPath`)
when generating error messages.  This should address issue #79.

### DIFF
--- a/conditional/params_test.go
+++ b/conditional/params_test.go
@@ -33,7 +33,7 @@ func TestIfMatch(t *testing.T) {
 	// Read request
 	r, _ := http.NewRequest(http.MethodGet, "https://example.com/resource", nil)
 	w := httptest.NewRecorder()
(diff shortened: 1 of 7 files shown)


mtiller:
@danielgtaylor Do you want me to reimplement this by fetching the `Router` from `Context`?

danielgtaylor:
@mtiller I think that would be great as it's a less intrusive change, thanks!

mtiller:
Is there anything you want me to do before merging this?  I'm still seeing issues related to schemas for generics that I'd like to fix.  But ideally I'd like to build that on top of this PR.

danielgtaylor (approved):
LGTM, thanks!

danielgtaylor:
Sorry, this got away from me during the holidays, thank you for the contribution! :+1:

PR #81, opened 2022-12-01
feat: schema handler can now handle schema names with special characters

mt35-rs:
Go types can have a `.` in them but the `chi` router can't handle this if the non-parameter part of the route starts with (contains?) a `.`.  Given all the potential other special characters in a Go type name (in the context of generics), it was easier to just use a wildcard and check for the proper suffix.

In the future, it would be pretty easy to add support for a `.yaml` prefix.

The error handler here was pretty bare bones.  It didn't respond with RFC 7807 problems, etc.  So I made a more consistent error writer.

This is all to address issue #74.

### DIFF
--- a/router.go
+++ b/router.go
@@ -352,12 +352,33 @@ func (r *Router) setupDocs() {
 		})
 	}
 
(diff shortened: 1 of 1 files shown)


danielgtaylor (approved):
Awesome, thanks!

PR #82, opened 2022-12-01
feat: ability to fetch router from huma context

mt35-rs:
This allows us to query the router for information.  I've added a GetOperation method so we can query for information about a given operation.  This is useful in hypermedia APIs because it allows us to link across different endpoints without having to hardwire paths.  Instead, the cross-referencing is done via operation id.

### DIFF
--- a/router.go
+++ b/router.go
@@ -25,11 +25,23 @@ var connContextKey contextKey = "huma-request-conn"
 // has finished.
 var opIDContextKey contextKey = "huma-operation-id"
 
+// routerContextKey is used to get the router associated with the API
+var routerContextKey contextKey = "huma-router"
+
 // GetConn gets the underlying `net.Conn` from a context.
 func GetConn(ctx context.Context) net.Conn {
 	conn := ctx.Value(connContextKey)
 	if conn != nil {
 		return conn.(net.Conn)
+	}
+	return nil
+}
+
+// GetRouter gets the `*Router` handling API requests
+func GetRouter(ctx context.Context) *Router {
+	router := ctx.Value(routerContextKey)
+	if router != nil {
+		return router.(*Router)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


router.go
```diff
@@ -532,7 +581,10 @@ func New(docs, version string) *Router {
 		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 			// Inject the operation info before other middleware so that the later
 			// middleware will have access to it.
-			req = req.WithContext(context.WithValue(req.Context(), opIDContextKey, &OperationInfo{}))
+			reqContext := req.Context()
+			withOpenID := context.WithValue(reqContext, opIDContextKey, &OperationInfo{})
+			withRouter := context.WithValue(withOpenID, routerContextKey, r)
+			req = req.WithContext(withRouter)
```
danielgtaylor, line 587:
```suggestion
			reqContext := req.Context()
			withOpID := context.WithValue(reqContext, opIDContextKey, &OperationInfo{})
			withRouter := context.WithValue(withOpID, routerContextKey, r)
			req = req.WithContext(withRouter)
```

router.go
```diff
@@ -243,6 +255,43 @@ func (r *Router) Resource(path string) *Resource {
 	return res
 }
 
+// GetOperation returns the path and
```
danielgtaylor, line 258: Missing the end of the sentence? :smile:

router.go
```diff
@@ -34,6 +37,15 @@ func GetConn(ctx context.Context) net.Conn {
 	return nil
 }
 
+// GetRouter gets the `*Router` handling API requests
+func GetRouter(ctx context.Context) *Router {
```
danielgtaylor, line 41: Can this be used to greatly simplify the code in #80? The context would already have the router injected and there's no need to pass it around everywhere right? 🤔

danielgtaylor (commented):
Looks good, just a couple minor things!

router.go
mt35-rs: Possibly.  Note that several tests don't have a router (in which case I send in `nil`).  But since the `GetRouter` function returns `nil` when it can't find a `Router` in the context, then it should actually be pretty equivalent.  I agree, it would simplify the code.  The issue here was that I wanted to provide you with independent PRs in case you decided to reject any.  So that's why I don't leverage this in #80.  But if you want me to refactor it once this PR is merged, I can certainly do that.

router.go
mt35-rs: This is fine for me.  Note that I was just following the naming convention of the context key, `opIDContextKey`.  So this change is fine but I'd say that for consistency you should then rename `opIDContextKey` to `openIDContextKey`.

router.go
mt35-rs: Addressed in [26405ef](https://github.com/danielgtaylor/huma/pull/82/commits/26405ef6270739c8356d65ee22dcbfbf9ffe69ca)

danielgtaylor (approved):
Thanks!

PR #86, opened 2023-01-05
fix: properly instantiate nested schemas

mt35-rs:
My previous fix for recursion prevented the crash but it wasn't a complete fix.  The issue I uncovered is that the OpenAPI specification data was incomplete.  You can see this by trying to access `/schemas/SomeSchema.json` where `SomeSchema` refers to a Go type that is not a response type but is nested inside a response type.  Before this fix, that would result in an error.  But just as bad, the OpenAPI spec generated (_i.e.,_ the contents of `/openapi.json`) were incomplete...schema definitions were missing there as well.

The reason for this is that my fix for infinite recursion was to simply replace all struct fields with a `$ref` to the actual schema rather than elaborate the schema in place.  As far as I can tell, this works fine.  But the reason it is incomplete is that the way `huma` constructs the OpenAPI components and their schemas is to simply generate a schema for the response type and publish that.  It didn't publish the nested schemas.  The reason is that previously _it didn't need to_ because they were all inline and so they would get picked up anyway.  But now they are not inline, they are simply referenced.  But that `$ref` pointer didn't resolve properly because it pointed to a schema definition that was never actually inserted into the OpenAPI specification document.

This change fixes that.  The fix is pretty simple and it arose from me not understanding how `huma` traverses the resources to build the OpenAPI spec.  With this fix in place, every time we skip over a nested schema `$ref` (to avoid recursion) while generate a schema, we make a note of this.  This information is then available after the schema is generated.  In the case of injecting the response schema into the OpenAPI components, we include a second pass where we _also_ inject any nested schemas if they haven't already been injected.

As a result, both the `/schemas/SomeType.json` URLs resolve to an actual JSON Schema document but also the `/openapi.json` file is complete (_i.e.,_ includes all definitions referenced via `$ref`).

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -97,8 +97,27 @@ func (c *oaComponents) addSchema(t reflect.Type, mode schema.Mode, hint string, 
 		// TODO: See if this type has a predefined schema associated with it
 		// via a predefined map, a callback or an interface.
 		var err error
-		if s, err = schema.GenerateWithMode(t, mode, nil, map[string]string{}); err != nil {
+		nestedSchemas := map[string]schema.NestedSchemaReference{}
+		if s, err = schema.GenerateWithMode(t, mode, nil, nestedSchemas); err != nil {
 			panic(err)
+		}
+		// Iterate over any nested schemas that are referenced by the schema
+		// we just generated
+		for k, v := range nestedSchemas {
+			// Check to see if this component already has a definition for that
+			// type in its schema collection
+			_, exists := c.Schemas[k]
+			if !exists {
+				// If not, generate the schema for the nested schema.
+				predefined := map[string]schema.NestedSchemaReference{}
+				s, err := schema.GenerateWithMode(v.Type, mode, nil, predefined)
+				if err != nil {
+					panic(err)
+				}
+				// Add the generated schema to the list of schemas associatd
+				// with this component.
+				c.Schemas[k] = s
+			}
 		}
 	}
 
--- a/operation.go
+++ b/operation.go
@@ -277,7 +277,8 @@ func (o *Operation) Run(handler interface{}) {
 				o.requests[ct].model = f.Type
 
 				if !o.requests[ct].override {
-					s, err := schema.GenerateWithMode(f.Type, schema.ModeWrite, nil, map[string]string{})
+					nestedSchemas := map[string]schema.NestedSchemaReference{}
+					s, err := schema.GenerateWithMode(f.Type, schema.ModeWrite, nil, nestedSchemas)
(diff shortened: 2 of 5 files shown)


danielgtaylor:
@mt35-rs I tried this out and while it generally makes sense and seems to work it does break the tests (I need to figure out why those don't run for this PR) and produces a bunch of weird changes for a complex production API of mine too, like:

```diff
@@ -25214,7 +31876,7 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/GetPreviewStreamsResponse"
+                  "$ref": "#/components/schemas/GetPreviewStreamsResponse2"
                 }
               }
             },
@@ -25609,7 +32271,7 @@
             "content": {
               "application/problem+json": {
                 "schema": {
-                  "$ref": "#/components/schemas/ErrorModel"
+                  "$ref": "#/components/schemas/ErrorModel2"
                 }
               }
             },
```

While seemingly innocent these have the potential to break generated SDKs and maybe other tools, so I'd like to get that fixed before merging. Otherwise this looks great! :+1:

mt35-rs:
See [here](https://github.com/danielgtaylor/huma/issues/87#issuecomment-1374931431) for some explanation of what is going on.

mt35-rs:
That is my hypothesis at least.  I haven't confirmed this.

openapi.go
```diff
@@ -97,9 +97,28 @@ func (c *oaComponents) addSchema(t reflect.Type, mode schema.Mode, hint string,
...
+			if !exists {
+				// If not, generate the schema for the nested schema.
+				predefined := map[string]schema.NestedSchemaReference{}
+				s, err := schema.GenerateWithMode(v.Type, mode, nil, predefined)
+				if err != nil {
+					panic(err)
+				}
+				// Add the generated schema to the list of schemas associatd
+				// with this component.
+				c.Schemas[k] = s
```
danielgtaylor, line 119: Found the fix. There is logic in `addExistingSchema` that wasn't being run, causing a later `DeepEqual` to fail and generating extra models. With this change the tests pass again for me.

```suggestion
				c.addExistingSchema(s, k, generateSchemaField)
```

danielgtaylor: @mt35-rs let me know if you want to apply this, otherwise I can open a new PR with the fix.

mt35-rs: Probably better for you to apply this.  You know the ins and outs better than I do.

danielgtaylor:
Fixed by #100. Note that you now need to opt-in to the new behavior via `schema.GenerateInline = false`

PR #97, opened 2023-03-17
feat: add server variables for server URL template substitution

GreyXor:
Allow to use server variables like this
image

ref: https://swagger.io/docs/specification/api-host-and-base-path/

Example PoC usage

```go
humaRouter.ServerLink("Example server", "{scheme}://{host}:{port}/api", map[string]huma.CaServerVariable{
	"scheme": {
		Enum:        []string{"http", "https"},
		Default:     "https",
		Description: "Example description 🔥",
	},
	"host": {
		Default:     "127.0.0.1",
		Description: "IP or domain where your server is located.",
	},
	"port": {
		Default:     "1337",
		Description: "Server port.",
	},
})
```

### DIFF
--- a/examples/bookstore/bookstore.go
+++ b/examples/bookstore/bookstore.go
@@ -60,7 +60,7 @@ var memoryDB = sync.Map{}
 func main() {
 	// Create a new router and give our API a title and version.
 	app := cli.NewRouter("BookStore API", "1.0.0")
-	app.ServerLink("Development server", "http://localhost:8888")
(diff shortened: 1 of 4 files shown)


danielgtaylor:
Thanks for the contribution and sorry for not merging sooner. This is now available in the Huma v2 beta as you get full access to the OpenAPI (including setting your own extension fields as needed). See https://github.com/danielgtaylor/huma/blob/de19e4c8484862ad678ce1bc544bf1867f2ef141/openapi.go#L70-L100

PR #119, opened 2023-09-05
Add GetMultipartForm method to huma context

lordspinach:
There is currently no way to get multipart form from a huma context. I need this feature in the Resolve method because I get a multipart request in one of the endpoints. At the moment the only way to get the form is through reflection, which I would like to avoid, as I think other developers would like to avoid as well.
I also renamed adapters struct methods receivers to 'c'. At the moment the adapters had 'c' and 'ctx' mixed in.

### DIFF
--- a/adapters/humachi/humachi.go
+++ b/adapters/humachi/humachi.go
@@ -3,6 +3,7 @@ package humachi
 import (
 	"context"
 	"io"
+	"mime/multipart"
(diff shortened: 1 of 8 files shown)


danielgtaylor (approved):

danielgtaylor:
Awesome, thank you!

PR #126, opened 2023-09-08
feat: add convenience model validator

### LINKED ISSUE #21: Support mark as required for query parameters
Currently, the `query` parameters are optional in the generated openapi docuemntation. However, some parameters are required in the API, and they are not suitable for `path`.

It would be great if huma could support **required** for query parameters.

### LINKED ISSUE #121: Validating JSON outside of a request
The validation is great, and it would be nice to be able to use it outside the context of a request (e.g., JSON file validation). However, the public interface is somewhat awkward due to the included optimizations.
Would it be in the spirit of things to add a simplified interface for generic schema validation? Or does supporting such a feature go out of the scope of Huma?

A few details that could potentially be abstracted away:
- The `Registry` parameter is used only if the schema is a reference (or is a parent to one, recursively)
- The `PathBuffer` could potentially be made optional
- `ValidateMode` could have a default (or be ignored?)
- `ValidateResult` could implement `Error()` and be returned as an `error`

I feel like a lot of the extra details could be abstracted away without giving up the optimizations.
For example, a config/option struct could be used that provides reasonable defaults.

Another approach might be to create a wrapper function that simplifies use (or even just suggest that users create one themselves).

danielgtaylor:
Adds a basic convenience utility for validating data against model schemas outside of the normal request/response flow.

Fixes #121.

### DIFF
--- a/README.md
+++ b/README.md
@@ -918,6 +918,28 @@ o.Extensions["x-cli-autoconfig"] = huma.AutoConfig{
 
 See the [CLI AutoConfiguration](https://rest.sh/#/openapi?id=autoconfiguration) documentation for more info, including how to ask the user for custom parameters.
 
+## Model Validation
+
+Huma includes a utility to make it a little easier to validate models outside of the normal HTTP request/response flow, for example on app startup to load example or default data and verify it is correct. This is just a thin wrapper around the built-in validation functionality, but abstracts away some of the boilerplate required for efficient operation and provides a simple API.
+
+```go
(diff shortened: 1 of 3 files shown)


danielgtaylor:
Thanks, that's an interesting approach to references and I like that schemas would contain all the information you need for e.g. validation. I'm going to merge this PR and will take a look at implementing your idea to see how it changes the code & interface because I can't visualize exactly how it may improve things without diving into the actual changes.

PR #131, opened 2023-09-19
feat: oneOf, anyOf, allOf, not schema support

### LINKED ISSUE #128: Schema: anyOf, allOf, oneOf, not
The OpenAPI 3.1 inherits the definition of a Schema Object from JSON Schema Draft 2020-12, which defines [anyOf, allOf, oneOf, and not](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-10.2.1)

All things considered, extending Validate and the Schema struct to work with this doesn't seem hugely invasive (albeit nontrivial).
What I see being most tricky is having a single endpoint that can accept more than one of bool, int, float, string, []SomeStruct and SomeStruct (my own use case involves a POST endpoint accepting both []SomeStruct and SomeStruct). Perhaps it would be possible to do something along the lines of what sse.Register does with the eventTypeMap?

danielgtaylor:
Adds support to the schema and validation for `oneOf`, `anyOf`, `allOf`, and `not` schemas. An example is included how these combined with a raw body can be used to create an operation with multiple input types.

Lays the groundwork to fix #128, though I want to revisit validation error messages and see how struct field tags could make use of these new features (particularly one-of support).

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -365,14 +365,12 @@ func Register[I, O any](api API, op Operation, handler func(context.Context, *I)
 	}
 	inputParams := findParams(registry, &op, inputType)
 	inputBodyIndex := -1
-	var inSchema *Schema
 	if f, ok := inputType.FieldByName("Body"); ok {
 		inputBodyIndex = f.Index[0]
-		inSchema = registry.Schema(f.Type, true, getHint(inputType, f.Name, op.OperationID+"Request"))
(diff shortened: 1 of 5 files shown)


nickajacks1:
I agree, this lays the groundwork for future additions, and it takes care of my use case beautifully.  Thanks!

PR #139, opened 2023-10-17
Fix: Update transforms.go to use explicit Host function from huma.Context

james-andrewsmith:
Hey @danielgtaylor 👋🏼 

Firstly - just found Huma and loving it - very quickly I've been able to get a TRPC-like workflow running which is amazing.

It looks like Chi does not provide the Host header through the Header func but it is explicitly available via the Host func. This meant the `$schema` never had a hostname and would return `https:///schema...`.

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -119,7 +119,7 @@ func (t *SchemaLinkTransformer) Transform(ctx Context, status string, v any) (an
 		return v, nil
 	}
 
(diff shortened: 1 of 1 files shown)


danielgtaylor (approved):
This was on my short list of things to fix ASAP but you beat me to it, thank you! 🚀

PR #192, opened 2023-12-26
Support array query parameters of primitives other than string

### LINKED ISSUE #191: Unsupported param type []int on query parameters
I am getting a panic when a request with a query parameter that is defined as an int array is sent.

ross96D:
fix #191

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -426,6 +426,20 @@ func transformAndWrite(api API, ctx Context, status int, ct string, body any) {
 	}
 }
 
+func parseArrElement[T any](values []string, parse func(string) (T, error)) ([]T, error) {
+	result := make([]T, 0, len(values))
+
+	for i := 0; i < len(values); i++ {
+		v, err := parse(values[i])
+		if err != nil {
+			return nil, err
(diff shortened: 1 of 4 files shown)


danielgtaylor:
@ross96D thanks and sorry for the delay in reviewing this PR. I've been busy with family for the holidays. This is a great addition but I have noticed a couple things:

1. Tests/coverage are not running on this PR. I need to fix that in the Github workflows ASAP.
2. Running tests/coverage manually a lot of this code isn't covered. I'm going to try and see how easy it is to change that as I'd prefer to not have coverage drop below 90%. There's a lot of additional error checking here which makes this kind of a pain.

Anyway, you're welcome to help with any of this, but I wanted to let you know that I'm also happy to do it and will try to get this merged within the next few days.

ross96D:
@danielgtaylor i have added some test and remove the support for query parameters with type []uint8 as this is treated as a byte array encoded with base64. The coverage is at 93.1%.

danielgtaylor:
@ross96D awesome thank you!

Issue #189, opened 2023-12-23
Response Body can not be recognised in other directory and package


qdongxu:
Versions:
1.  github.com/danielgtaylor/huma/v2 v2.1.0 
2.  go 1.21.0

The `my-app` in the tutorial works fine.

after expand `my-app`, adding code all in main.go ( the **`main`** package), the open API spec works as expected[demo code 1](https://github.com/qdongxu/huma-my-app/tree/workwell):

Response Example:
```json
{
  "$schema": "http://example.com",
  "code": 0,
  "data": {
    "Greeting": "string",
    "Name": "string"
  },
  "msg": "string"
}
```

but after moving the expanded code to other directory and  package ( the **`my-app/module/submodule`** package), the open API spec pages shows errors [demo code 2](https://github.com/qdongxu/huma-my-app/tree/main) :

The `Responses` Section shows

>  Body   application/json
> No schema defined
```
Resonse Example:

Example cannot be created for this schema
Error: Invalid reference token: Responsemy-app
```

This is the first time I use huma. Is it a bug or any convention I did not follow? thanks for a kindly clarification.

danielgtaylor:
Fascinating! Using a struct from another package *does* work as expected, but the combination of generics + a struct in another package is an edge case I haven't run into before. Rather than returning just the name of the type, I get something like `Response[my-app/module/submodule.BizData]` from a `reflect.Type.Name()` which gets turned into an invalid JSON pointer reference, which is why you are getting the error. Sorry about that.

I will need to update the [DefaultSchemaNamer](https://github.com/danielgtaylor/huma/blob/main/registry.go#L22-L40) to take into account generics with types from other modules. For now to get unstuck you could try [passing a modified schema namer function](https://github.com/danielgtaylor/huma/blob/main/defaults.go#L69) until this is fixed in Huma.

PR #199, opened 2024-01-04
fix ref name contains recursive brackets '[]'

qdongxu:
PR #195 fixed ISSUE #189. But regex matched a shorter string when a generic parameter is an slice( that means there's embedded brackets). 

And the expression is more complex when the generic parameter is another generic type. 

Propose to use a full path name as  the name is used as an ID and reference in the yaml. Readability is trivial.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ mem.out
 docs/site
 docs/__pycache__
 docs/.cache
+.idea/
--- a/registry.go
+++ b/registry.go
@@ -4,13 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
-	"regexp"
 	"strings"
 )
-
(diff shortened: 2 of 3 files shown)


danielgtaylor:
@qdongxu thanks, this is unfortunately more complex than I had originally realized. I like your proposal but disagree that the readability of the naming is trivial because these names get directly used in generated SDKs as class/struct names. If at all possible I *really* want to keep them nice, so I created #200. Please check that out and see if it covers all your use cases and I'll merge and push a new release ASAP.

qdongxu:
@danielgtaylor thanks for you quick fix. The new solution works perfect. And I got a bonus from the [v2.2.0](https://github.com/danielgtaylor/huma/releases/tag/v2.2.0)  release: array from the query parameter. thanks for such a quick release.

PR #208, opened 2024-01-18
fix: avoid sending referrer header to unpkg

costela:
Sending the Referrer header is unnecessary and may leak some private data (unlikely to be highly-sensitive, but I think the point is still valid).


danielgtaylor (approved):
Thanks! :+1:

PR #209, opened 2024-01-18
proposal: switch to versioned spotlight and add integrity check

costela:
Using an unversioned import can become a security issue in case of compromised npm repositories or compromised unpkg.
To make the situation a bit safer-by-default, I propose a switch to versioned imports + the use of [subresource integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) to ensure users have a comparable level of security for the JS side as they do for the go side.

IMHO the trade-off between "bleeding-edge" and "security" seems to fall in favor of security, especially since uses who want the bleeding-edge can still override the docs handler.

WDYT?

### DIFF
--- a/api.go
+++ b/api.go
@@ -367,8 +367,9 @@ func NewAPI(config Config, a Adapter) API {
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
     <title>Elements in HTML</title>
     <!-- Embed elements Elements via Web Component -->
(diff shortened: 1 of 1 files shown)


danielgtaylor:
@costela yeah this is a great idea. This is marked as a WIP, so let me know if you have additional work to do or are ready for it to be merged :+1:

costela:
@danielgtaylor now we're good to go! Was just waiting for #208 to avoid a rebase :sweat_smile:

danielgtaylor (approved):

PR #212, opened 2024-01-20
feat: add echo framework adapter

x-user:
Add support for labstack/echo framework (echo.HandlerFunc).


danielgtaylor (approved):
Awesome, thank you! 🚀

PR #225, opened 2024-02-02
Errors in OAuth2 how to examples

x-user:
Fix minor errors in OAuth2 how to, but huma.WriteErr requires huma.API as first argument and there no way to obtain it in middleware. I was thinking about replacing it with http.Error, but it requires http.ResponseWriter and huma.Context doesn't implement it. So i'm stuck there.

### DIFF
--- a/docs/docs/how-to/oauth2-jwt.md
+++ b/docs/docs/how-to/oauth2-jwt.md
@@ -147,7 +147,7 @@ func NewJWKSet(jwkUrl string) jwk.Set {
 
 // NewAuthMiddleware creates a middleware that will authorize requests based on
 // the required scopes for the operation.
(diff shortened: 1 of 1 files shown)


x-user:
I can use
```go
ctx.SetStatus(http.StatusUnauthorized)
ctx.BodyWriter().Write([]byte(http.StatusText(http.StatusUnauthorized)))
```
But it looks kinda ugly.

danielgtaylor:
@x-user thanks for fixing this up! You could maybe pass the `api` in when creating the middleware, then you get content negotiation and the normal error model in the error response:

```go
api.UseMiddleware(NewAuthMiddleware(api, "https://example.com/.well-known/jwks.json"))
```

x-user:
Yes closures can solve this. Then i will make it this way.

danielgtaylor (approved):
Great, thank you! 🚀

PR #237, opened 2024-02-10
Support parsing default option value formats

ddl-ebrown:
- This simple change allows for specifying a format struct tag in CLI options to influence how string -> primitive type conversion works.

   This is not exhaustive and only supports time.Duration parsing right now (since time.Duration is really an int64 but often is useful to represent using a time string like `5s`)

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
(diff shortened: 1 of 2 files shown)


danielgtaylor:
@ddl-ebrown note that I didn't just close this PR, your commit was merged as part of #245 and will be in the next release. Thanks! 👍🏻

PR #239, opened 2024-02-11
docs: adds echo router to BYOR page

certanet:
Adds Echo to BYOR page and moves FIber up to restore *alphabetical order*


danielgtaylor (approved):
Thanks! :+1:

Issue #241, opened 2024-02-14
Best way to cancel a request from a middleware?


martin31821:
Hi,

thanks for the awesome piece of software, it's been a really nice ride so far, but I've stumbled upon a problem I haven't been able to solve yet:

I'm trying to implement a simple jwt validation middleware and I wanted to cancel (i.e. not call the operation handler) from my middleware, but there seems to be no way to indicate an error in the middleware chain.

How would I best do this? 
Can I just not call "next" and write a body?
How would I then generate the nice error objects?

danielgtaylor:
@martin31821 that's right, you just skip calling `next(ctx)` if you want the middleware to terminate the request processing chain.

You can use the [`huma.WriteErr`](https://pkg.go.dev/github.com/danielgtaylor/huma/v2#WriteErr) utility function for returning an error, which will use the API's registered error model and do content negotiation with the registered formats. Here's a quick example to trigger via query param:

```go
api.UseMiddleware(func(ctx huma.Context, next func(ctx huma.Context)) {
	if ctx.Query("error") == "true" {
		huma.WriteErr(api, ctx, http.StatusInternalServerError, "Some friendly message", fmt.Errorf("error detail"))
		return
	}

	next(ctx)
})
```

You can also use [`huma.ErrorDetail`](https://pkg.go.dev/github.com/danielgtaylor/huma/v2#ErrorDetail) instead of `fmt.Errorf(...)` to provide more information to the client when appropriate.

Using the [greeting example API](https://huma.rocks/tutorial/your-first-api/#operation), you would trigger this by going to http://localhost:8888/greeting/world?error=true and see a response like this:

```yaml
$ restish :8888/greeting/world?error=true
HTTP/1.1 500 Internal Server Error
Content-Length: 154
Content-Type: application/problem+cbor
Date: Thu, 15 Feb 2024 05:17:01 GMT

{
  $schema: "http://localhost:8888/schemas/ErrorModel.json"
  detail: "Some friendly message"
  errors: [
    {
      message: "error detail"
    }
  ]
  status: 500
  title: "Internal Server Error"
}
```

martin31821:
perfect, thanks. Didn't find the helper method mentioned anywhere in the docs in that context :)

danielgtaylor:
> Didn't find the helper method mentioned anywhere in the docs in that context :)

@martin31821 thanks for keeping me honest about this. :blush: :joy: I added some docs in #248!

PR #248, opened 2024-02-16
docs: add doc about middleware errors

danielgtaylor:
This adds a small section to the docs for returning nice standardized well structured errors that utilize the API's content negotiation and marshaling from a router-agnostic middleware. Thanks to #241 for pointing out this was missing!

### DIFF
--- a/docs/docs/features/middleware.md
+++ b/docs/docs/features/middleware.md
@@ -73,9 +73,35 @@ func NewHumaAPI() huma.API {
 }
 ```
 
+### Errors
+
+If your middleware encounters an error, you can stop the processing of the next middleware or operation handler by skipping the call to `next` and writing an error response.
+
+The [`huma.WriteErr(api, ctx, status, message, ...error)`](https://pkg.go.dev/github.com/danielgtaylor/huma/v2#WriteErr) function can be used to write nice structured error responses which respect client-driven content negotiation for marshaling:
+
+```go title="code.go"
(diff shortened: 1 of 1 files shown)


docs/docs/features/middleware.md
```diff
@@ -73,9 +73,35 @@ func NewHumaAPI() huma.API {
...
+
+If your middleware encounters an error, you can stop the processing of the next middleware or operation handler by skipping the call to `next` and writing an error response.
+
+The [`huma.WriteErr(api, ctx, status, message, ...error)`](https://pkg.go.dev/github.com/danielgtaylor/huma/v2#WriteErr) function can be used to write nice structured error responses which respect client-driven content negotiation for marshaling:
+
+```go title="code.go"
+func MyMiddleware(ctx huma.Context, next func(ctx huma.Context)) {
+	// If there is a query parameter "error=true", then return an error
+	if ctx.Query("error") == "true" {
+		huma.WriteErr(api, ctx, http.StatusInternalServerError,
```
ferdefiore, line 86: @danielgtaylor sorry, one question here. Where the `api` is supposed to be defined? I'm trying to create a middleware in a separate file so I'm not using

```
api.UseMiddleware(func(ctx huma.Context, next func(ctx huma.Context)) {
	[code with access to api]
})
```

Also how can I send parameters to the middleware? for example I have a `keysRetriever` app that I need to use inside the middleware. Thanks!

Issue #253, opened 2024-02-18
Access huma.context inside the register handler


spa5k:
There needs to be a way to manually handle the huma.Context, and not just through the Resolve function inside the handler itself to set headers, etc.

Otherwise it gets really complex and weird trying to set something like cookie

```go
huma.Register(api, huma.Operation{
		OperationID: "list-things",
		Method:      http.MethodGet,
		Path:        "/things",
		Summary:     "Get a filtered list of things",
	}, func(ctx context.Context, input *MyInput) (*MyOutput, error) {
		fmt.Printf("Host: %s\n", input.Host)
		fmt.Printf("Name: %s\n", input.Name)
		// Set the cookie
		cookie := &http.Cookie{
			Name:  "state",
			Value: "abcd",
		}
		cookieStr := "Set-Cookie: " + cookie.String()
		return &MyOutput{ID: "abc", State: "111", CookieState: cookieStr}, nil
	})
	```
	
	This is just my experiment on somehow set the cookie, still failing on it.

**Comment by danielgtaylor** [comment]
@spa5k you can define a `Set-Header` header in the response struct and then set it to the string value of the cookie. Your example is close:

```go
// In your response struct:
type DemoOutput struct {
	SetCookie string `header:"Set-Cookie"`
}

// In your operation handler:
resp := &DemoOutput{}
c := http.Cookie{
	Name:   "foo",
	Value:  "bar",
	MaxAge: 1000,
}
resp.SetCookie = c.String()
```

https://go.dev/play/p/mOmh8TxeKC2

See also https://huma.rocks/features/response-outputs/#headers

Unfortunately reading cookies is not so easy right now, and we are using #227 to track improving that, as well as making sure you can set multiple cookies in a single response. Those features will be coming soon and I'm actively working on it.

> There needs to be a way to manually handle the huma.Context, and not just through the Resolve function inside the handler itself to set headers, etc.

Part of the philosophy of Huma is that all your inputs/outputs are declared and strongly typed in order to prevent surprises and magic that makes maintenance difficult and can introduce hard to find bugs over time. Access to `huma.Context` directly in resolvers is meant to be an escape hatch when needed, but it's preferable to use the declarative mechanism of setting headers through the response type when possible. It's an adjustment in behavior as most frameworks let you set whatever you want whenever you want however you want - Huma adds some guardrails earned through many hard learned lessons that have hurt our customers and engineering teams over the last few years.

danielgtaylor:
[...]

PR #258, opened 2024-02-22
feat: better support for cookies

### LINKED ISSUE #227: Best practices for managing cookies
Hi there, thanks for such an amazing open source project. I've only just discovered Huma and am loving using it so far!

I'm wondering what the best practice is for accessing and setting cookies on requests and responses respectively? My use-case is around managing authentication sessions; I'd like to be able to access a cooke in a Huma middleware, if possible, but I'm unable to find an obvious utility to do so.

I suspect it's possible to go directly to the cookie header using something like `ctx.Header("Cookie")` from the Huma context, but I was hoping there might be a more convenient mechanism similar to other frameworks?

Thanks!

ETA: I'd also be interested to hear about any strategies for setting cookies from within request handlers?

danielgtaylor:
This PR does the following:

1. Adds a `cookie` field tag for cookie parameters. Fields can be any of the normal parameter types, plus `http.Cookie` is supported to get the entire parsed cookie.
2. A new function `huma.ReadCookies(ctx huma.Context, filter string)` is provided for reading/parsing cookies from one or more `Cookie` headers in a router-agnostic way and without needing to modify the `huma.Context` interface. This function can be used in middleware to read cookie values.
3. If a response header type supports `fmt.Stringer` (aka a `String() string` method), then that is used to serialize the header.
4. If a response header type is a slice, then `huma.Context().AppendHeader(...)` is used to write multiple headers with the same name.
5. As a result of 3 and 4 above, `http.Cookie` and `[]http.Cookie` are supported for response headers using the `Set-Cookie` header. They basically just work.
6. There are examples and docs for each scenario. See the diff for details.

Fixes #227, #253.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,3 +2,4 @@ ignore:
   - benchmark
   - examples
   - adapters
+  - cookie.go
--- a/cookie.go
+++ b/cookie.go
@@ -1,21 +1,36 @@
 package huma
 
 import (
+	"fmt"
 	"net/http"
 	"net/textproto"
 	"strings"
 )
 
-// ReadCookies reads all cookies from the request headers. If a filter is
-// provided, only cookies with that name are parsed & returned.
-func ReadCookies(ctx Context, filter string) []*http.Cookie {
+// ReadCookie reads a single cookie from the request headers by name. If
+// multiple cookies with the same name exist, the first is returned.
+func ReadCookie(ctx Context, name string) (*http.Cookie, error) {
 	headers := []string{}
 	ctx.EachHeader(func(name, value string) {
 		if strings.EqualFold(name, "cookie") {
 			headers = append(headers, value)
 		}
 	})
-	return readCookies(headers, filter)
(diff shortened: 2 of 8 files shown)


x-user:
Looks good for me only minor annoyance in needing to extract single cookie from slice when using `hume.ReadCookies` and filter kinda strange name to get single result.

danielgtaylor:
@x-user this was just copied from the standard library, see https://github.com/golang/go/blob/master/src/net/http/cookie.go#L273-L310. I'm open to changing our exported function signature though.

x-user:
@danielgtaylor i understand this, but standard library uses different public signature
```go
func (r *Request) Cookies() []*Cookie
func (r *Request) Cookie(name string) (*Cookie, error)
```

danielgtaylor:
@x-user take a look at the latest and let me know what you think.

x-user:
@danielgtaylor, i have no complaints.

PR #270, opened 2024-02-29
Added UUID as a type for path/query/body

jh125486:
I was using this to create a demo app today and noticed that I couldn't use `uuid.UUID` as a path param... so here's a PR I hope you'll accept.

- Added UUID as a schema type.
- Moved "special types" up above the big type switch [`schema.go#564`]... I think that should be OK?
- Migrated deprecated `PtrTo()` -> `PointerTo()`.
- Updated `go.mod`.
- Fixed linter warnings.
- Tiny grammar fixes.


victoraugustolls:
Is it possible to make the `uuid` type itself customizable? I question this because at my company, we use this [package](https://github.com/gofrs/uuid) due to the support for UUIDv7.

jh125486:
Also v6/v7 are experimental, they are implemented in the Google package: uuid package - github.com/google/uuid - Go Packages<https://pkg.go.dev/github.com/google/uuid#NewV7>

V/r
Jacob Hochstetler

From: Victor ***@***.***>
Date: Thursday, February 29, 2024 at 19:10
To: danielgtaylor/huma ***@***.***>
Cc: Jacob Hochstetler ***@***.***>, Author ***@***.***>
Subject: Re: [danielgtaylor/huma] Added UUID as a type for path/query/body (PR #270)

Is it possible to make the uuid type itself customizable? I question this because at my company, we use this package<https://github.com/gofrs/uuid> due to the support for UUIDv7.

danielgtaylor:
@jh125486 thanks for the PR! I see a lot of great changes here :+1:

Thinking this through a bit I believe your approach is reasonable, but I'd really like to [avoid additional dependencies](https://dgt.hashnode.dev/reducing-go-dependencies) trying to support various new types people might want to use as params. I think we can work around this by instead relying on the stdlib's [encoding.TextUnmarshaler](https://pkg.go.dev/encoding#TextUnmarshaler) interface, so any type which implements that method would be automatically supported.

When combined with [custom field schemas](https://huma.rocks/features/schema-customization/#field-schema) you should be able to do something like this and just have it work as Google's `uuid` package [implements `UnmarshalText`](https://pkg.go.dev/github.com/google/uuid#UUID.UnmarshalText):

```go
import "github.com/google/uuid"

type UUID struct {
	uuid.UUID
}

func (u *UUID) Schema(r huma.Registry) *huma.Schema {
	return &huma.Schema{Type: huma.TypeString, Format: "uuid"}
}

	OperationID: "demo",
	Method:      http.MethodGet,
	Path:        "/demo/{example}",
	Example UUID `path:"example"`
}) (*GreetingResponse, error) {
	// ...
})
```

Full example (which fails currently as it needs `TextUnmarshaler` support added): https://go.dev/play/p/5M4Xf0ilq-d

Feel free to take a stab at updating the code to support that interface, otherwise I can do it when I have some time soon. I appreciate the time and effort you put into this!

jh125486:
@danielgtaylor, 
Pushed a commit, and I think this is what you're looking for.

danielgtaylor:
@jh125486 thanks! Your commits are merged in #276, I just had to fix the build as it was using an older version of golangci-lint in the Github actions. Thanks! :+1:

PR #271, opened 2024-03-01
fix: Set body height to 100% of the viewport height

mailbaoer:
Thank you for the fantastic work your team has done. When I use huma for a demo, I noticed that the documentation does not display correctly on my extended screen, as follows:
image

I found a solution on the Stoplight documentation website that suggests resolving the issue by adding a style configuration to the body.
```html
<body style="height: 100vh;">
    <stoplight-api />
</body>
```
[Stoplight Layout](https://docs.stoplight.io/docs/elements/d536b917a4e1d-layout)

Hope this issue can be fixed!


danielgtaylor (approved):
Thanks! :+1:

PR #275, opened 2024-03-02
feat: add request context method

gmtstephane:
## Add WithContext Adapter method 

Hello ✋🏻 ,

In most popular golang routers, there is a way to modify the request context in order to put some middlewares values.
In my use case, this would be to set user context, so that in the auth routes, i can retrieve the user.

### Example with echo 

```golang
func UserCookieMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
	return func(c echo.Context) error {

		sessionCookie, err := c.Cookie(Name)
		if err != nil {
			slog.Error("cookie error:", err)
		}

		user, err := Parse(sessionCookie.Value)
		if err != nil {
			slog.Error("cookie error:", err)
		}

		c.SetRequest(c.Request().WithContext(WithUser(c.Request().Context(), user)))
		return next(c)
	}

}

```

### Proposition 

Simply add a WithContext method to the `huma.Context` , that will set the inner context from adapters 

```golang
func AuthCookieMiddlware(ctx huma.Context, next func(huma.Context)) {
	sessionCookie, err := huma.ReadCookie(ctx, Name)
	if err != nil {
		slog.Error("cookie error:", err)
	}

	user, err := Parse(sessionCookie.Value)
	if err != nil {
		slog.Error("cookie error:", err)
	}

	ctx.WithContext(WithUser(ctx.Context(), user))

	next(ctx)

}

```

I would try to avoid using the base router since your framework handle already a lot of stuff. 
It would be better to use only huma.

Thanks for this framework, is is really well documented.

If you have features in mind, feel free to message me .

### DIFF
--- a/api.go
+++ b/api.go
@@ -371,7 +371,7 @@ func NewAPI(config Config, a Adapter) API {
             integrity="sha256-yIhuSFMJJ6mp2XTUAb4SiSYneP3Qav8Uu+7NBhGJW5A="
             crossorigin="anonymous"></script>
   </head>
-  <body>
+  <body style="height: 100vh;">
 
     <elements-api
       apiDescriptionUrl="` + config.OpenAPIPath + `.yaml"


danielgtaylor:
@gmtstephane this probably isn't well documented, but I assumed you could wrap the context with your own overrides like this:

```go
type humaContext huma.Context

type MyContext struct {
	humaContext
	custom context.Context
}

func (c *MyContext) Context() context.Context {
	return c.custom
}
```

Then later use it in middleware like this:

```go
api.UseMiddleware(func(ctx huma.Context, next func(huma.Context)) {
	// Make an updated context available to the next middleware.
	next(&MyContext{
		humaContext: ctx,
		custom:      context.WithValue(ctx.Context(), "foo", "bar"),
	})
})

api.UseMiddleware(func(ctx huma.Context, next func(huma.Context)) {
	fmt.Println("Getting from context:", ctx.Context().Value("foo"))
	next(ctx)
})
```

Here is a runnable example: https://go.dev/play/p/mm00vcbmsPT

I'm wondering if there is a use-case for actually updating the underlying request context like in your proposed code that wouldn't be covered by the wrapping functionality above. What do you think?

gmtstephane:
Yes , but the use case would be to access the context value in the handler
function , not in another middleware.

The middleware function is taking a huma.context where the handler function
is taking a generic context.Context.



Le dim. 3 mars 2024 à 08:31, Daniel G. Taylor ***@***.***> a
écrit :

Insei:
I would prefer that huma did not give the opportunity to get into the depths of routers and http-server, etc., because Huma provides an additional level of abstraction, which allows you to avoid dragging things from Adapters to a higher level, this both narrows the level of capabilities and simplifies development. And also, if we abandon abstraction levels, it will be more difficult to control what is happening, various errors and conflicts may appear. I would like only Huma to be able to manage the routers context when processing requests.

But this is just my opinion, but it is precisely because of these qualities that I choose Huma in my projects. This framework allows me to focus on designing APIs, and not on routers, error handling, describing OpenAPI specs, etc.

But I agree, many basic things are not currently implemented in Router Agnostic middlewares and there is a temptation to use some of the capabilities that routers provide, but my idea is to write similar middlewares for huma and eliminate the temptation.

danielgtaylor:
@Insei I was torn about what to do here, but I think I agree. Let's keep the request abstracted & immutable until we have a discrete use-case that requires us to change that behavior.

@gmtstephane while this PR looks great, I'm going to say let's not change the `huma.Context` interface. You are correct that this is a common scenario, so here's what I propose:

1. Add a `huma.WithValue(ctx huma.Context, key string, value any) huma.Context` (or maybe `huma.WithContext`? Then you could also set timeouts/deadlines?) Either way this can use the sample code I provided above to wrap the context and then it's easy for middleware authors to use.
2. Add some documentation under the middleware and maybe request resolvers sections of the docs to describe how this is used so people know it exists and don't have to build their own.

What do you all think?

gmtstephane:
I am good with that. 

I'll do that when i have a bit of time ahead this week

leonklingele:
> Add a `huma.WithValue(ctx huma.Context, key string, value any) huma.Context`

Please make it a `huma.WithValue(ctx huma.Context, key, value any) huma.Context` so we can set "private values type using an unexported key type as such:

```go
type contextKeyType string

const (
    requestIDContextKey contextKeyType = "request_id"
)

// …

huma.WithValue(ctx, requestIDContextKey, "the-request-id)
```

This prevents others with access to the context to retrieve back the value.

api.go
```diff
@@ -63,6 +63,9 @@ type Context interface {
 	// Context returns the underlying request context.
 	Context() context.Context
 
+	//WithContext sets the underlying request context.
```
leonklingele, line 66:
```suggestion
	// WithContext sets the underlying request context.
```

x-user:
> > Add a `huma.WithValue(ctx huma.Context, key string, value any) huma.Context`
> 
> Please make it a `huma.WithValue(ctx huma.Context, key, value any) huma.Context` so we can set "secret" values using an unexported key type as such:
> 
> ```go
> type contextKeyType string
> 
> const (
>     requestIDContextKey contextKeyType = "request_id"
> )
> 
> // …
> 
> huma.WithValue(ctx, requestIDContextKey, "the-request-id")
> ```
> 
> This prevents others with access to the context to retrieve back the value.

Thus PR isn't merged already because there existing way to do this ([link](https://github.com/danielgtaylor/huma/pull/275#issuecomment-1975073339)).

PR #278, opened 2024-03-04
fix: allow response types to contain unexported fields

FlorianLoch:
So far using a type as `Body` on a response type containing unexported fields results in a panic:

```
panic: reflect.StructOf: field "config" is anonymous but has PkgPath set

goroutine 1 [running]:
reflect.runtimeStructField({{0x105753038, 0x6}, {0x105778f34, 0x27}, {0x1059311a8, 0x1058b15e0}, {0x10575303f, 0x8}, 0x0, {0x14000352668, ...}, ...})
        /opt/homebrew/Cellar/go/1.22.0/libexec/src/reflect/type.go:2538 +0x1c8
reflect.StructOf({0x14000270008, 0x10, 0x14000548508?})
        /opt/homebrew/Cellar/go/1.22.0/libexec/src/reflect/type.go:2189 +0x274
github.com/danielgtaylor/huma/v2.(*SchemaLinkTransformer).OnAddOperation(0x140005a2150, 0x140005aa0b0, 0x14000548601?)
       .../huma/transforms.go:110 +0x2d0
```

As I am using `ent` (entgo.io) that is subpar, because the structs generated by `ent` contain unexported fields (in this case `config`, which triggered the panic above).
Creating proxy types would really break the experience, so it's not really an option. 
And anyway I don't see why `huma` would not want to support such structs as it seems to me there is no drawback in this simple fix suggested - but as I am new to `huma` and its codebase I might just not see it 😉.

Feedback would be appreciated.

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -97,10 +97,11 @@ func (t *SchemaLinkTransformer) OnAddOperation(oapi *OpenAPI, op *Operation) {
 			}
 			for i := 0; i < typ.NumField(); i++ {
 				f := typ.Field(i)
-				fields = append(fields, f)
 				if f.IsExported() {
-					// Track which fields are exported so we can copy them over. It's
-					// preferred to track/compute this here to avoid allocations in
+					fields = append(fields, f)
+
+					// Track which fields are exported, so we can copy them over.
+					// It's preferred to track/compute this here to avoid allocations in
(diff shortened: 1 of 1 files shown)


danielgtaylor (approved):
This looks great, thanks! :+1:

PR #284, opened 2024-03-05
feat: add convenience functions for summary and description

xarunoba:
This extends the convenience functions further by generating phrases and sentences for `Summary` and `Description` respectively.

I didn't want to add further dependencies like `cases` and whatnot since I'm not sure if you're fine with that or not. This looks a bit hacky but it does the trick for a basic implementation.

Tell me what you think :)

### DIFF
--- a/docs/docs/features/operations.md
+++ b/docs/docs/features/operations.md
@@ -71,7 +71,7 @@ huma.Get(api, "/things/{thing-id}", func(ctx context.Context, input *YourInput) 
 })
 ```
 
-The generated operation ID for the above example would be `get-things-by-thing-id`. Override `huma.GenerateOperationID(method, path string, response any)` to customize the generated operation IDs.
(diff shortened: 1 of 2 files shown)


danielgtaylor:
@xarunoba thanks for the PR! I think adding the summary is useful since it'll show up in the docs instead of just the URL, and it's sort of reasonable as that's how you would describe the URL given no other info. The description I'm not so sure about though - what does this give us? It seems to just be the same thing without adding any value. I think if people want more than a generated ID + summary then they should just use `huma.Register` for full control. What do you think?

I'm happy to merge if you remove the description and then document the generated summary behavior and how to override it, see https://github.com/danielgtaylor/huma/blob/main/docs/docs/features/operations.md#convenience-methods for where to edit. Thanks!

xarunoba:
@danielgtaylor hmm I agree with your standpoint. I just included description for the sake of why not (aha)
I have removed the description and added the documentation about the generated summary behavior and how to override it. :)

danielgtaylor (approved):
Thanks! :+1:

Issue #290, opened 2024-03-07
Don't indirect-require certain modules


leonklingele:
I prefer to keep the `go.mod` file to clean as possible.

Since I migrated from `net/http` to Huma with https://github.com/helsinki-systems/settleup/commit/8d68a4dcae113f03d98ca6fec1e4e70c05332961, the `go.mod` now indirectly requires a few modules which imo it should not, e.g.:

```
	github.com/inconshreveable/mousetrap v1.1.0 // indirect
	github.com/spf13/cobra v1.8.0 // indirect
	github.com/spf13/pflag v1.0.5 // indirect
```
(See https://github.com/helsinki-systems/settleup/blob/c193e84673b94642136e81b917cd36f0109feef5/go.mod#L14-L16)

I couldn't really see why those dependencies are required. Would it be possible to remove them somehow?

danielgtaylor:
@leonklingele funny you should mention that... I recently wrote a post on [Reducing Go Dependencies](https://dgt.hashnode.dev/reducing-go-dependencies) and have been thinking about how best to approach Cobra. The problem is that removing it will be a breaking change, but what I'm thinking (and I have local code I need to clean up for a PR) is to move the [CLI](https://huma.rocks/features/cli/) functionality into a `humacli` package, then mark the originals as deprecated for a few releases before removing them. It'll be breaking, but with enough lead time it should be acceptable and I'm happy to consider it a "fix".

One common thing we use the CLI for is [adding a quick command to dump the OpenAPI](https://github.com/danielgtaylor/huma/blob/main/examples/spec-cmd/main.go#L68-L75), but it can also be used for DB management, migrations, convenience commands to do other stuff, etc.

danielgtaylor:
For anyone following this issue, I have created #291 to create the new package and mark the existing calls as deprecated.

danielgtaylor:
The new `humacli` package is now released in https://github.com/danielgtaylor/huma/releases/tag/v2.8.0. It'll be a while before I remove the old functionality as it's a breaking change, so I'm going to close this issue as we are well on our way and you can now build without the old code using a tag. Thanks for the help all!

PR #291, opened 2024-03-09
fix: move CLI to its own package, deprecate huma.NewCLI

danielgtaylor:
This PR moves the built-in optional CLI functionality into its own `humacli` package and updates all examples & docs. The existing CLI functionality stays but is marked as deprecated to give people time to move to the new package. Switching is fairly easy:

- `huma.NewCLI` → `humacli.New`
- `huma.CLI` → `humacli.CLI`
- `huma.Hooks` → `humacli.Hooks`
- `huma.WithOptions` → `humacli.WithOptions`

The plan is to **remove the old functionality a few releases from now**. This will be a small breaking change to facilitate the removal of some dependencies that are not used by everyone, but will be treated as a bug fix.

Related to #290. This doesn't fix it yet, but sets us on the track to doing so soon.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -78,6 +78,6 @@ func WithOptions[Options any](f func(cmd *cobra.Command, args []string, options 
 //
 // Deprecated: use `humacli.New` instead.
 func NewCLI[O any](onParsed func(Hooks, *O)) CLI {
-	log.Print("huma.NewCLI is deprecated, use humacli.New instead")
+	log.Println("huma.NewCLI is deprecated, use humacli.New instead")
 	return humacli.New(onParsed)
 }
--- a/cli.go
+++ b/cli.go
@@ -1,3 +1,6 @@
+//go:build !humanewclipackage
+// +build !humanewclipackage
+
 package huma
 
 import (
--- a/cli.go
+++ b/cli.go
@@ -32,7 +32,7 @@ type Hooks = humacli.Hooks
 //
 // Deprecated: use `humacli.WithOptions` instead.
 func WithOptions[Options any](f func(cmd *cobra.Command, args []string, options *Options)) func(*cobra.Command, []string) {
-	log.Print("huma.WithOptions is deprecated, use humacli.WithOptions instead")
+	log.Println("huma.WithOptions is deprecated, use humacli.WithOptions instead")
 	return humacli.WithOptions(f)
 }
 
--- a/cli.go
+++ b/cli.go
@@ -1,18 +1,9 @@
 package huma
 
 import (
-	"context"
-	"fmt"
-	"os"
-	"os/signal"
-	"path/filepath"
-	"reflect"
-	"strconv"
-	"strings"
-	"syscall"
-	"time"
+	"log"
 
-	"github.com/danielgtaylor/casing"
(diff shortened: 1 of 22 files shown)


cli.go
```diff
@@ -61,156 +29,11 @@ var durationType = reflect.TypeOf((*time.Duration)(nil)).Elem()
...
-				if err != nil {
-					panic(err)
-				}
-			}
-			flags.BoolP(name, field.Tag.Get("short"), def, field.Tag.Get("doc"))
-		default:
-			panic("Unsupported option type: " + field.Type.Kind().String())
-		}
-	}
+	log.Print("huma.WithOptions is deprecated, use humacli.WithOptions instead")
```
leonklingele, line 35:
```suggestion
	log.Println("huma.WithOptions is deprecated, use humacli.WithOptions instead")
```

cli.go
```diff
@@ -252,40 +75,9 @@ func (c *cli[O]) setupOptions(t reflect.Type, path []int) {
...
-		case <-quit:
-			if c.stop != nil {
-				fmt.Println("Gracefully shutting down the server...")
-				c.stop()
-			}
-		}
-
-	}
-	return c
+	log.Print("huma.NewCLI is deprecated, use humacli.New instead")
```
leonklingele, line 81:
```suggestion
	log.Println("huma.NewCLI is deprecated, use humacli.New instead")
```

humacli/humacli.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+
+		name := field.Tag.Get("name")
+		if name == "" {
+			name = casing.Kebab(field.Name)
+		}
+
+		envName := "SERVICE_" + casing.Snake(name, strings.ToUpper)
+		defaultValue := field.Tag.Get("default")
+		if v := os.Getenv(envName); v != "" {
```
x-user, line 189: I think `os.LookupEnv` is better choice. Because with it we can override default value with empty string if needed.
```go
if v, ok := os.LookupEnv(envName); ok {
```

leonklingele:
I wonder if we could use a build constraint to already prevent building with the legacy code:

// +build !humanewclipackage

and

go build -tags humanewclipackage

simonfr:
Hi,
Sorry but it's a bit strange : you choose to "deprecate" `huma.NewCLI` but you just delete it from the old package. Looks like a breaking no ?

danielgtaylor:
@simonfr thanks for the feedback. Please take a look at https://github.com/danielgtaylor/huma/releases/tag/v2.13.0

The `huma.NewCLI` call was deprecated for several releases with a note about migrating to the new `humacli` package, giving users some time to migrate before removal. While removing it in v2.13.0 is technically a breaking change, it is treated as a fix because it is the only way to remove the underlying hard dependency requirement from the codebase for users who do not want to use the CLI functionality. It turns out this is important for security purposes and in certain organizations like government agencies and NGOs with extremely strict dependency and auditing requirements.

This wasn't a decision taken lightly due to the potential of upsetting people with a small breaking change, however the call was made that it was worth it to enable more people to use Huma. Even if you disagree with this change, I hope you can understand why it was done.

PR #292, opened 2024-03-09
Add WithContext function

gmtstephane:
Hello, 

I accidentally closed the last pull request #275 . 

Are you talking about something like this in order to set the context ?

### DIFF
--- a/api.go
+++ b/api.go
@@ -36,8 +36,10 @@ type ResolverWithPath interface {
 	Resolve(ctx Context, prefix *PathBuffer) []error
 }
 
-var resolverType = reflect.TypeOf((*Resolver)(nil)).Elem()
-var resolverWithPathType = reflect.TypeOf((*ResolverWithPath)(nil)).Elem()
+var (
+	resolverType         = reflect.TypeOf((*Resolver)(nil)).Elem()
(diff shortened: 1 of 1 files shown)


x-user:
I think @danielgtaylor mean that it should be done on middleware level:

```go
// wrapper.go

type (
	humaContext huma.Context

	wrapper struct {
		humaContext
		custom context.Context
	}
)

var wrapperPool = sync.Pool{
	New: func() interface{} {
		return &wrapper{}
	},
}

func acquireWrapper() *wrapper {
	return wrapperPool.Get().(*wrapper)
}

func (w *wrapper) Context() context.Context {
	return w.custom
}

// middleware.go

type (
	contextKey struct{}

	HumaMiddleware func(huma.Context, func(huma.Context))
)

const {
	requestIDKey = contextKey{}
}

func Get(c huma.Context) string {
	ctx := c.Context()
	v := ctx.Value(requestIDKey)
	if v == nil {
		return ""
	} else {
		return v.(string)
	}
}

func New() HumaMiddleware {
	return func(c huma.Context, next func(huma.Context)) {
		w := acquireWrapper()
		w.humaContext = c
		w.custom = context.WithValue(c.Context(), requestIDKey, "request-id")
		defer wrapperPool.Put(w)

		next(w)
	}
}
```

PR #298, opened 2024-03-12
fix: length validation for utf8 strings

maximdanilchenko:
Hi! Current string length validation made with go's standard `len` function works incorrectly with non ascii characters. So if there any non-english symbols `"maxLength"` validator will fail when it shouldn't.

In this PR I'v replaced `len` with `utf8.RuneCountInString`. Instead of number of bytes it will calculate number of runes. 

It will resolve most issues. But it don't take into account [unicode text segmentation](https://unicode.org/reports/tr29/). So if there will be symbols such as emojis, that consists of more than one unicode character it still will calculate length incorrectly. More info could be found [here](https://stackoverflow.com/a/12668840). 

So may be the next step is to add usage of https://github.com/rivo/uniseg. But considering it is a big change and new dependancy may be it is not worth it.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 	"unsafe"
 )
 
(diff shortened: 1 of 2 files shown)


danielgtaylor:
@maximdanilchenko thanks for the PR! This one is interesting to me because the length may be used to limit storage in e.g. a DB, so I wonder if we should be counting bytes, runes, or visible characters. It might be useful to give users the choice but that might also be overkill. I think counting runes is probably a good default.

x-user:
IMHO: it's used to validate input string length so it should count runes.

danielgtaylor (approved):

Issue #296, opened 2024-03-12
Possibilit of `dependentRequired` tag


victoraugustolls:
Is it possible to implement the `[dependentRequired](https://json-schema.org/draft/2020-12/json-schema-validation#name-dependentrequired)` validation tag? It is part of the json-schema 2020-12 that OpenAPI 3.1 supports!

danielgtaylor:
@victoraugustolls yes I'd be open to adding support for it, and am happy to review a PR for it. It sounds like you'll want something like:

1. Update [`huma.Schema`](https://github.com/danielgtaylor/huma/blob/main/schema.go#L58) to add something like `DependentRequired map[string][]string`. Maybe add some pre-computed error messages for validation.
2. Update the [map validation](https://github.com/danielgtaylor/huma/blob/main/validate.go#L526-L674) to check if that new field is set in the schema, and if so look up which other fields are required as you iterate through each map entry.
3. Decide how it should be used, either by manually creating schemas or if there should be a high-level way to enable it using struct field tags. If struct tags, how would they work? Happy to discuss that here so we have a plan.
4. Add tests to keep coverage up.
5. Update any relevant [docs](https://github.com/danielgtaylor/huma/tree/main/docs/docs)

Thanks!

victoraugustolls:
@danielgtaylor great! I will work on a pull request for it.

victoraugustolls:
@danielgtaylor closing this issue!

PR #303, opened 2024-03-12
feat: dependent required tag

victoraugustolls:
This pull request adds the ability to define some fields to be required **only** when other fields are present with the struct tag `dependentRequired`.

Resolves issue #296


victoraugustolls:
@danielgtaylor do you think we should write a `how-to` docs for `conditional-fields`?

victoraugustolls:
@danielgtaylor while doing this pull request I noticed that there was a `schema.Required` and `schema.requiredMap`, so I followed the same approach for the `dependentRequired`. But we could get away only with the `DependentRequired` if that is okay with you. I will push the changes and let me know if you want me to revert them.

danielgtaylor (approved):
LGTM, great work! Thank you :+1:

I think it would be good to panic if there is a typo or the wrong field name is provided, so I'll hold off merging unless you want to do that in another PR.

danielgtaylor:
Oh, and one last thing, you'll want to modify https://github.com/danielgtaylor/huma/blob/main/docs/mkdocs.yml#L54-L59 so your new how-to doc shows up in the menu properly!

victoraugustolls:
@danielgtaylor done! added the panic for non existing fields and updated the docs!

danielgtaylor (approved):

PR #306, opened 2024-03-13
feat: breaks infinite cycle when looking for fields in recursive structures.

yuridevx:
My machine has crashed when I've tried to use Huma for the new micro-service that I'm developing.
```
// Node is a custom type for testing recursive definition for huma.Register
type Node struct {
	Name  string          `json:"name"`
	Nodes []Node          `json:"nodes,omitempty"`
	Left  *Node           `json:"left,omitempty"`
	Named map[string]Node `json:"named,omitempty"`
}
```
This pr breaks the cycle in findInType by adding `visited map[reflect.Type]bool` as a parameter.
I haven't spent much time figuring out semantics and usage of findInType, so it might be a wrong solution. However I would like to hear your thoughts on the issue.

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -370,11 +370,11 @@ func (r *findResult[T]) EveryPB(pb *PathBuffer, v reflect.Value, f func(reflect.
 
 func findInType[T comparable](t reflect.Type, onType func(reflect.Type, []int) T, onField func(reflect.StructField, []int) T, recurseFields bool, ignore ...string) *findResult[T] {
 	result := &findResult[T]{}
-	_findInType(t, []int{}, result, onType, onField, recurseFields, ignore...)
+	_findInType(t, []int{}, result, onType, onField, recurseFields, make(map[reflect.Type]bool), ignore...)
 	return result
 }
 
-func _findInType[T comparable](t reflect.Type, path []int, result *findResult[T], onType func(reflect.Type, []int) T, onField func(reflect.StructField, []int) T, recurseFields bool, ignore ...string) {
+func _findInType[T comparable](t reflect.Type, path []int, result *findResult[T], onType func(reflect.Type, []int) T, onField func(reflect.StructField, []int) T, recurseFields bool, visited map[reflect.Type]bool, ignore ...string) {
 	t = deref(t)
 	zero := reflect.Zero(reflect.TypeOf((*T)(nil)).Elem()).Interface()
 
@@ -392,6 +392,10 @@ func _findInType[T comparable](t reflect.Type, path []int, result *findResult[T]
 
 	switch t.Kind() {
 	case reflect.Struct:
+		if visited[t] {
+			return
+		}
+		visited[t] = true
 		for i := 0; i < t.NumField(); i++ {
 			f := t.Field(i)
 			if !f.IsExported() {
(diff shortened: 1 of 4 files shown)


huma.go
```diff
@@ -370,11 +370,11 @@ func (r *findResult[T]) EveryPB(pb *PathBuffer, v reflect.Value, f func(reflect.
 
 func findInType[T comparable](t reflect.Type, onType func(reflect.Type, []int) T, onField func(reflect.StructField, []int) T, recurseFields bool, ignore ...string) *findResult[T] {
 	result := &findResult[T]{}
-	_findInType(t, []int{}, result, onType, onField, recurseFields, ignore...)
+	_findInType(t, []int{}, result, onType, onField, recurseFields, make(map[reflect.Type]bool), ignore...)
 	return result
 }
 
-func _findInType[T comparable](t reflect.Type, path []int, result *findResult[T], onType func(reflect.Type, []int) T, onField func(reflect.StructField, []int) T, recurseFields bool, ignore ...string) {
+func _findInType[T comparable](t reflect.Type, path []int, result *findResult[T], onType func(reflect.Type, []int) T, onField func(reflect.StructField, []int) T, recurseFields bool, visited map[reflect.Type]bool, ignore ...string) {
```
victoraugustolls, line 377: Does the visited map need to be a boolean? By the looks of it, it only matters if a key exists and not the boolean value, but correct me if I'm wrong! In the case that I'm not, what about:

```suggestion
func _findInType[T comparable](t reflect.Type, path []int, result *findResult[T], onType func(reflect.Type, []int) T, onField func(reflect.StructField, []int) T, recurseFields bool, visited map[reflect.Type]struct{}, ignore ...string) {
```

huma_test.go
```diff
@@ -45,6 +45,14 @@ type UUID struct {
 	uuid.UUID
 }
 
+// Node is a custom type for testing recursive definition for huma.Register
+type Node struct {
+	Name  string          `json:"name"`
+	Nodes []Node          `json:"nodes,omitempty"`
+	Left  *Node           `json:"left,omitempty"`
+	Named map[string]Node `json:"named,omitempty"`
+}
```
victoraugustolls, line 54: I know you took this from your example case! But what about using more explicit names like `RecusiveEmbedded`, following existing test type names?

huma_test.go
```diff
@@ -806,6 +814,28 @@ func TestFeatures(t *testing.T) {
 			URL:    "/round-trip",
 			Body:   `{"$schema": "...", "name": "foo"}`,
 		},
+		{
+			Name: "recursive schema",
```
victoraugustolls, line 818: I think there is also a need to check how the schema is created by creating a test case in the `TestSchema` test! Could you please create one? It is in the `schema_test.go` file.

victoraugustolls:
@danielgtaylor, sorry if I'm barging myself in here! I just guessed that reviewing this PR would be a nice opportunity to learn more about the code.

huma.go
yuridevx: Boolean are cleaner syntactically, plus if you benchmark it has 0 impact on performance.

yuridevx:
@danielgtaylor @victoraugustolls I've adjusted visited to struct{} to match the rest of project usage, also I've added break to getFields

danielgtaylor (approved):
@yuridevx thank you! :+1:

Issue #313, opened 2024-03-14
Nested structs visibility not working properly


CptKirk:
It seems that visibility struct tags like `readOnly` and `writeOnly` are not applied to nested structs. To help understanding the problem I created a small example test in `huma_test.go` in function `TestFeatures`:
```go
{
	Name: "request-body-nested-struct-readOnly",
	Register: func(t *testing.T, api huma.API) {
		type NestedStruct struct {
			Foo struct {
				Bar string `json:"bar"`
			} `json:"foo" readOnly:"true"`
			Value string `json:"value"`
		}
		huma.Register(api, huma.Operation{
			Method: http.MethodPost,
			Path:   "/body",
		}, func(ctx context.Context, input *struct {
			Body *NestedStruct
		}) (*struct{}, error) {
			assert.Equal(t, true, true)
			return nil, nil
		})
	},
	Method: http.MethodPost,
	URL:    "/body",
	Body:   `{"value":"test"}`,
},
```

I expect the test to pass but it fails due to this error:
```json
{
  "$schema":"https:///schemas/ErrorModel.json",
  "title":"Unprocessable Entity",
  "status":422,
  "detail":"validation failed",
  "errors":[
    {
      "message":"expected required property foo to be present",
      "location":"body",
      "value":{"value":"test"}
    }
  ]
}
```
It says that property `foo` is required but it is set to `readOnly="true"` so this error is not correct.

I also noticed, that the `writeOnly` struct tag does not work as I expect it to. I thought that that the `writeOnly` tag would remove the corresponding fields from the response, but they are still sent in the response, is this expected behavior?

victoraugustolls:
>  I thought that that the writeOnly tag would remove the corresponding fields from the response, but they are still sent in the response, is this expected behavior?

It might not be the same thing, but `writeOnly` is only set in the request; I don't think it would impact the response (at least, I wouldn't expect it to).

bekabaz:
I have an issue that is related to this, and when I use the most up to date main, it is still present. The only difference in mine, is that it is saying a property is required, but it is not.

I have this as my input for the Operation:

```
Body struct {
	CreateSeq
}
```
and CreateSeq is defined as:
```
type CreateSeq struct {
	Ci *CI `json:"ci"`
	Name          string         `json:"name" required:"true"`
	Status        string         `json:"status" enum:"active,inactive" default:"inactive"`
}
```

This is the error I get on a request:
```
...
"message": "expected required property ci to be present",
[...]

Issue #273, opened 2024-03-01
Default HTML title of Docs


elmarw:
How to change the default title of the html generate docs on /docs?

The default title seems to be Stoplight's "Elements in HTML".

danielgtaylor:
@elmarw the docs are set up here by default but can be disabled for customization:

https://github.com/danielgtaylor/huma/blob/main/api.go#L355-L386

You can just set `config.DocsPath = ""` and then add your own handler using whatever docs settings or tools you like:

```go
config := huma.DefaultConfig("My API", "1.0.0")
config.DocsPath = ""

api := humachi.New(router, config)
api.Adapter().Handle(&Operation{
	Method: http.MethodGet,
	Path:   config.DocsPath,
}, func(ctx Context) {
	ctx.SetHeader("Content-Type", "text/html")
	ctx.BodyWriter().Write([]byte(`whatever you want here!`))
})
```

You can just copy the default and modify the title, or use something entirely different like [SwaggerUI](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/installation.md#unpkg) or [Scalar Docs](https://github.com/scalar/scalar?tab=readme-ov-file#from-a-cdn). It's up to you!

danielgtaylor:
@elmarw hopefully this answers the question. If not, feel free to re-open the issue!

PR #314, opened 2024-03-14
Nested struct visibility

CptKirk:
This pull request enables `readOnly`, `writeOnly` and `deprecated` struct tags on nested structs.

Resolves issue: #313


danielgtaylor:
@CptKirk I added a commit with a slightly different approach. JSON Schema allows for `$ref` to have siblings for validation (see [spec example](https://json-schema.org/draft/2020-12/json-schema-core#name-collecting-annotations)), in this case we generate something like this:

```yaml
type: object
properties:
  foo:
    $ref: '#/components/schemas/FooSchema'
    readOnly: true
```

The validator didn't support that as it always resolved refs *before* checking validations. The commit I added changes that behavior to check at the top level, which supports how we generate schemas and works fine if no `$ref` is present. This has the added benefit of not requiring a new method in the `Registry` interface, and it doesn't run into issues when you re-use a struct which is sometimes read-only and sometimes not.

Please let me know what you think!

The only weird thing I ran into is that Stoplight Docs don't mark it as read-only, but that seems to be a bug as both [Swagger UI and Scalar Docs](https://github.com/danielgtaylor/huma/issues/273#issuecomment-1973618623) do the right thing for this simple example:

```go
type GreetingOutput struct {
	Body struct {
		Another struct {
			Foo string `json:"foo"`
		} `json:"another" readOnly:"true"`
	}
}
```

Screenshot 2024-03-14 at 22 39 32
Screenshot 2024-03-14 at 22 41 13

CptKirk:
@danielgtaylor thanks for providing the input! I checked out your solution and indeed I can confirm, that the validation works correctly with your changes but the stoplight docs do not. The problem seems to be, that Stoplight parses the OpenAPI spec differently than Swagger UI, because with my proposed changes Stoplight Docs become correct. I have created a little test case to figure out what the exact problem is. I created the following case:
```go
type NestedStruct struct {
	Value int `json:"value"`
	Foo   struct {
		Bar string `json:"bar"`
	} `json:"foo" readOnly:"true"`
}

func main() {
	router := chi.NewMux()

		OperationID: "post-nested",
		Method:      http.MethodPost,
		Path:        "/nested",
		Summary:     "Create a nested",
		Tags:        []string{"Nested"},
	}, func(
		ctx context.Context,
		input *struct {
			Body *NestedStruct
		}) (
		*struct {
			Body *NestedStruct
		}, error) {
		resp := &NestedStruct{}
		return &struct{ Body *NestedStruct }{Body: resp}, nil
	})

	http.ListenAndServe("127.0.0.1:8888", router)
}
```

Then I downloaded the OpenAPI spec for mine and your solution, ran a diff and indeed there is a small difference:
```diff
--- danielgtaylor.json	2024-03-15 11:51:57.546482363 +0100
+++ cptkirk.json	2024-03-15 11:52:40.267077258 +0100
@@ -100,20 +100,21 @@
           ],
           "type": "object"
         },
         "NestedStructFooStruct": {
           "additionalProperties": false,
           "properties": {
             "bar": {
               "type": "string"
             }
           },
+          "readOnly": true,
           "required": [
             "bar"
           ],
           "type": "object"
         }
       }
     },
     "info": {
       "title": "My API",
       "version": "1.0.0"
```

I agree that a new method in the `Registry` interface is not the ideal solution, but not having correct Stoplight docs would be a bummer for DX, at least in our case. Could we solve the problem another way so the `Schema` for the struct in the `Registry` has the right flags?

danielgtaylor:
@CptKirk the problem I'm running into is this scenario:

```go
type Foo struct {
	Field string `json:"field"`
}

type ExampleResponse1 struct {
	Body struct {
		Example Foo `json:"example" readOnly:"true"`
	}
}

type ExampleResponse2 struct {
	Body struct {
		Example Foo `json:"example"` // NOT readOnly!
	}
}
```

In this case it is incorrect to mark the generated `Foo` schema as `readOnly` as it is only *sometimes* read-only!

I wonder if we can get Stoplight to fix their support of `$ref` with other validation 🤔

CptKirk:
@danielgtaylor Ah, now I see the problem. Yeah you are right, this is a case were my proposed solution would break things. Thanks for taking the time explaining it to me. I think you can merge your solution as it fixes the problem with validation. I will look over the weekend into the issue with Stoplight and will try to fix it.

PR #323, opened 2024-03-18
 feat(validation): flexible validation messages for pattern tag

purin-dev:
Adds a new `patternDescription` struct tag to allow for human-readable error messages when using the pattern tag. 

Error messages are formatted like `expected string to be {description}` - this allows the user to use their own articles while still being consistent with the other validators


danielgtaylor (approved):
@purin-dev awesome feature; thank you! :+1:

PR #324, opened 2024-03-19
Multipart Form RawBody Type Support

jamelt:
I'm not entirely sure if this is the direction you want to go in. But this makes adding simple upload endpoints more approachable.

### DIFF
--- a/docs/docs/features/request-inputs.md
+++ b/docs/docs/features/request-inputs.md
@@ -58,7 +58,7 @@ The special struct field `Body` will be treated as the input request body and ca
 
 ### Other Body Types
 
-Sometimes, you want to bypass the normal body parsing and instead read the raw body conents directly. This is useful for unstructured data, file uploads, or other binary data. You can use `RawBody []byte` **without** a `Body` field to access the raw body bytes without any parsing/validation being applied. For example, to accept some `text/plain` input:
+Sometimes, you want to bypass the normal body parsing and instead read the raw body contents directly. This is useful for unstructured data, file uploads, or other binary data. You can use `RawBody []byte` **without** a `Body` field to access the raw body bytes without any parsing/validation being applied. For example, to accept some `text/plain` input:
 
 ```go title="code.go"
 huma.Register(api, huma.Operation{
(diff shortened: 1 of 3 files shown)


danielgtaylor (approved):
Sorry for the delay as I wasn't 100% sure what to do about multipart forms and how best to handle them. A couple thoughts:

1. This PR is great, thank you! I think allowing raw access to the form as a convenience is a good idea and will merge it in.
2. I think in the future it would be great to have a `huma.Format` that can unmarshal multipart forms into user-defined structs just like you unmarshal JSON, but that's for another PR. For now raw access to the form is a step in the right direction.

Thanks! 🚀

jamelt:
@danielgtaylor Hey, no worries. I know that life can get busy. I just wanted to put something in there. I thought I saw you put a comment in a ticket that you were open to folks contributing multipart form support. 

So, I tried to stick it in there while fitting nicely into the design spirit of what you had there. 

But your suggestion for number two is fantastic. I had that thought as well. But considering my requirement was uploading files, it seemed like quite a bit of a stretch to try to support a bunch of generic parameters and unmarshaling those into JSON. 

Also, the structure of the possibilities of the multipart form's key values seems quite flexible to the extent that you might have to make some choices on how the JSON mapping will work. 

But it would help because you could then get your open API spec to document what properties are available in these multipart form endpoints.

Thanks again!

JanRuettinger:
With this PR the docs now include the following example:

```
type MyInput struct {
	ID      string `path:"id"`
	Detail  bool   `query:"detail" doc:"Show full details"`
	Auth    string `header:"Authorization"`
	Body    MyBody
	RawBody []byte
}
```

That works. However, ideally the RawBody can also be of the type `multipart.Form`. If I change the input to the following:

```
type MyInput struct {
	Auth    string `header:"Authorization"`
	Body    MyBody
	RawBody multipart.Form
}
```

I get the warning: ` WARN  http: superfluous response.WriteHeader call from github.com/danielgtaylor/huma/v2/adapters/humachi.(*chiContext).SetStatus (humachi.go:86)`

Is there a way to use a multipart.Form RawBody and a structured Body at the same time?

minpeter:
I am in exactly the same situation.
i need help..

danielgtaylor:
@JanRuettinger @minpeter good catch. These should be fixed by #329.

PR #328, opened 2024-03-21
feat: hidden body field support

### LINKED ISSUE #327: hidden:"true" not working as expected
I could have sworn this was working prior, but I don't have any screenshots of it to prove it. I have a response struct:

```
type AccountResponse2 struct {
	Body Account2
}

type Account2 struct {
	Id        int64  `json:"id" doc:"The ID of the resource."`
	Protected *string `json:"protected,omitempty" hidden:"true"`
}
```

The docs show this as the model for AccountResponse2
image
I would have expected *Protected* to not show up at all.

It's totally possible I'm doing something wrong, but also could be a legit bug. Thoughts?

edit: i tried downgrading to v2.8.0 but it's the same thing so maybe this was always like this and I'm just noticing it

bekabaz:
Attempting to address https://github.com/danielgtaylor/huma/issues/327 which will allow the use of the `hidden` validation tag on Body fields. To avoid double checking of the hidden tag, I moved the existing check from findParams() to only be in SchemaFromField(). All existing tests are passing, however I am not familiar enough with the entire repository to guarantee correctness. Any feedback is welcome.

Fixes #327.

### DIFF
--- a/docs/docs/features/request-validation.md
+++ b/docs/docs/features/request-validation.md
@@ -15,38 +15,56 @@ type Person struct {
 }
 ```
 
-The standard `json` tag is supported and can be used to rename a field and mark fields as optional using `omitempty`. The following additional tags are supported on model fields:
+The standard `json` tag is supported and can be used to rename a field and mark fields as optional using `omitempty`. Any field tagged with `json:"-"` will be ignored in the schema.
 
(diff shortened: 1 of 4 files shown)


danielgtaylor (approved):
Thanks 👍🏻

PR #330, opened 2024-03-23
Allow specifying type aliases to provide schema definitions

lsdch:
This is meant to allow registering type aliases to be used when generating `Schema`. It is an alternative to implementing the [`SchemaProvider`](https://huma.rocks/features/schema-customization/#field-schema) interface, which is not always very convenient (e.g. when dealing with types from external packages which can not be directly extended).

Example usage: 
```go
registry := huma.NewMapRegistry("#/components/schemas", huma.DefaultSchemaNamer)

registry.RegisterTypeAlias(reflect.TypeFor[some_external_pkg.OptionalStr](), reflect.TypeFor[string]())
registry.RegisterTypeAlias(reflect.TypeFor[some_external_pkg.OptionalDateTime](), reflect.TypeFor[time.Time]())
```


lsdch:
Sorry, the test was using `reflect.TypeFor` which was only introduced in Go 1.22

lsdch:
I made sure the linter is happy this time, my local setup was using the `--fast` flag and did not detect this error

PR #305, opened 2024-03-13
feat(config): allow to specify an optional API prefix to play well with chi router mounting

burgesQ:
If the chi router attached to huma is mounted into another router (let say - to delimit the `/api` namespace), then the OpenAPI browser rendering is broken because of failure to fetch the served YAML documentation.

The MR add an extra `APIPrefix` to the huma.Config struct. The value is then prefixed to the documentation path in the template rendering.
  

```golang
rootMx, apiMx := chi.NewMux(), chi.NewMux()

prefix := "/api" 
rootMx.Use(RedirectToSwagger)
rootMx.Mount(prefix, apiMx)

cfg := huma.DefaultConfig("My API", "1.0.0")
cfg.APIPrefix = prefix

api := humachi.New(apiMx, cfg)

huma.Register(api, huma.Operation{
		OperationID: "get-greeting",
		Method:      http.MethodGet,
		Path:        "/greeting/{name}",
		Summary:     "Get a greeting",
		Description: "Get a greeting for a person by name.",
		Tags:        []string{"Greetings", "Utils"},
	}, rs.getGreeting)


return &http.Server{handler: rootMx}
```
```console
🗀 src/common/api/gopi/src  master-gopi-poc 🏷️  notag ※ 993237a +181 -471 *?!✘  via 🐹 v1.22 
11:58:14 ❯ http :4242/api/greeting/q                                                   🌿 zsh 🐏 (19%|6%)
HTTP/1.1 200 OK
Content-Encoding: gzip
Content-Length: 110
Content-Type: application/json
Date: Wed, 13 Mar 2024 10:58:20 GMT
Link: </schemas/GreetingOutputBody.json>; rel="describedBy"
Vary: Accept-Encoding

{
    "$schema": "http://localhost:4242/schemas/GreetingOutputBody.json",
    "message": "Hello, q!"
}
```

PR #338, opened 2024-03-28
feat: support OpenAPI servers base path & router groups

### LINKED ISSUE #331: Streamlining Base Path Setup in OpenAPI Schemas
Currently, configuring base paths in OpenAPI schemas is a bit cumbersome. Simplifying this process by allowing direct usage of base paths like localhost:8008/api/v1 would greatly enhance user experience. It not only makes usage easier but also generates more concise code names, avoiding unwieldy ones like apiV1AuthGoogleCallbackGet.

https://swagger.io/docs/specification/api-host-and-base-path/

I suggest implementing a way to define the base path schema as the option while generating the huma instance.

danielgtaylor:
This PR is an alternative approach to #305 by @burgesQ to enable serving the OpenAPI and schemas at the correct path when given a router group / sub-router at some base path like `/api`. It works like this:

```go
mux := chi.NewMux()
mux.Route("/api", func(r chi.Router) {
	config := huma.DefaultConfig("My API", "1.0.0")
	config.Servers = []*huma.Server{{URL: "https://example.com/api"}}
	api = humachi.New(r, config)

	// TODO: register operations
	huma.Get(api, "/demo", ...)
})
http.ListenAndServe("localhost:8888", mux)
```

This creates a sub-router at `/api` and then sets the `https://example.com/api` server URL including the base path. The sub-router is used to create the Huma instance. The `/demo` route registered via Huma then becomes `/api/huma` when the service is running, and all the docs/openapi/schemas/etc will be linked correctly.

I personally like the idea of using the already existing OpenAPI server base path functionality rather than duplicating it via a prefix/basePath parameter in the config. What do you think?

@spa5k, I believe this should fix #331. Please let me know if this would work for you.

I also did a quick example of how other routers can do this, e.g. for Gin you can use `humagin.NewWithGroup(router, group, config)` (you have to pass both the router and group since `*gin.RouterGroup` has no `ServeHTTP` method).

For the Go 1.22+ `http.ServeMux` via `humago` there is now a prefix you can set which acts similarly to the other router's groups. Use `humago.NewWithPrefix(mux, prefix, config)`.

This should also enable us to support rewriting API gateways, e.g. document the server's URL with base path like `https://example.com/api` but then do *not* use a sub-router as the incoming request to Go will never see the `/api` which was stripped off. Any other cases I'm missing?

Fixes #305, #331.

### DIFF
--- a/adapters/adapters_test.go
+++ b/adapters/adapters_test.go
@@ -87,21 +87,23 @@ func testAdapter(t *testing.T, api huma.API) {
 }
 
 func TestAdapters(t *testing.T) {
-	config := huma.DefaultConfig("Test", "1.0.0")
+	config := func() huma.Config {
+		return huma.DefaultConfig("Test", "1.0.0")
+	}
 
 	for _, adapter := range []struct {
 		name string
 		new  func() huma.API
 	}{
-		{"chi", func() huma.API { return humachi.New(chi.NewMux(), config) }},
-		{"chi4", func() huma.API { return humachi.NewV4(chi4.NewMux(), config) }},
-		{"echo", func() huma.API { return humaecho.New(echo.New(), config) }},
-		{"fiber", func() huma.API { return humafiber.New(fiber.New(), config) }},
-		{"gin", func() huma.API { return humagin.New(gin.New(), config) }},
-		{"httprouter", func() huma.API { return humahttprouter.New(httprouter.New(), config) }},
-		{"mux", func() huma.API { return humamux.New(mux.NewRouter(), config) }},
-		{"bunrouter", func() huma.API { return humabunrouter.New(bunrouter.New(), config) }},
-		{"bunroutercompat", func() huma.API { return humabunrouter.NewCompat(bunrouter.New().Compat(), config) }},
+		{"chi", func() huma.API { return humachi.New(chi.NewMux(), config()) }},
+		{"chi4", func() huma.API { return humachi.NewV4(chi4.NewMux(), config()) }},
+		{"echo", func() huma.API { return humaecho.New(echo.New(), config()) }},
+		{"fiber", func() huma.API { return humafiber.New(fiber.New(), config()) }},
+		{"gin", func() huma.API { return humagin.New(gin.New(), config()) }},
+		{"httprouter", func() huma.API { return humahttprouter.New(httprouter.New(), config()) }},
+		{"mux", func() huma.API { return humamux.New(mux.NewRouter(), config()) }},
+		{"bunrouter", func() huma.API { return humabunrouter.New(bunrouter.New(), config()) }},
+		{"bunroutercompat", func() huma.API { return humabunrouter.NewCompat(bunrouter.New().Compat(), config()) }},
 	} {
 		t.Run(adapter.name, func(t *testing.T) {
 			testAdapter(t, adapter.new())
--- a/adapters/humaecho/humaecho.go
+++ b/adapters/humaecho/humaecho.go
@@ -91,8 +91,13 @@ func (c *echoCtx) BodyWriter() io.Writer {
 	return c.orig.Response()
 }
 
+type router interface {
+	Add(method, path string, handler echo.HandlerFunc, middlewares ...echo.MiddlewareFunc) *echo.Route
+}
+
 type echoAdapter struct {
-	router *echo.Echo
+	http.Handler
+	router router
 }
 
 func (a *echoAdapter) Handle(op *huma.Operation, handler func(huma.Context)) {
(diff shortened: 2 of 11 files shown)


x-user:
Not tested but should work.
```go
type Router interface {
	Add(string, string, echo.HandlerFunc, ...echo.MiddlewareFunc) *echo.Route
}

type echoAdapter struct {
	http.Handler
	router Router
}

// ...

func New(r *echo.Echo, config huma.Config) huma.API {
	return huma.NewAPI(config, &echoAdapter{Handler: r, router: r})
}

// NewWithGroup creates a new Huma API using the provided Echo router and group,
// letting you mount the API at a sub-path. Can be used in combination with
// the `OpenAPI.Servers` field to set the correct base URL for the API / docs
// / schemas / etc.
func NewWithGroup(r *echo.Echo, g *echo.Group, config huma.Config) huma.API {
	return huma.NewAPI(config, &echoAdapter{Handler: r, router: g})
}
```

danielgtaylor:
@x-user I adapted your suggestions to the latest commit. Unfortunately the Fiber adaptor doesn't actually implement the underlying request connection's `SetReadDeadline` and results in a nil pointer panic, so I've kept the `router.Test(r)` call for now. May be able to fix this upstream but it'll take a bit to be released.

x-user:
> @x-user I adapted your suggestions to the latest commit. Unfortunately the Fiber adaptor doesn't actually implement the underlying request connection's `SetReadDeadline` and results in a nil pointer panic, so I've kept the `router.Test(r)` call for now. May be able to fix this upstream but it'll take a bit to be released.

It's happens because `net.Conn` embedded in `fasthttp.fakeAddrer` is nil.

burgesQ:
Sadly, this approach won't cover my use-case where I need to have 2 routers each with their own middlewares: 

```golang 
rootMx, apiMx := chi.NewMux(), chi.NewMux()

rootMx.Use(httplog.RequestLogger(logger))
apiMx.Use(middleware.JWT)
apiMx.NotFound(jsonNotFound)
rootMx.Mount("/api", apiMx)

cfg := huma.DefaultConfig("My API", "1.0.0")

cfg.Prefix = "/api" // missing

api := humachi.New(apiMx, cfg)

huma.AutoRegister(api, struct{}{})

huma.AutoRegister()
```

x-user:
> Sadly, this approach won't cover my use-case where I need to have 2 routers each with their own middlewares:
> 
> ```go
> rootMx, apiMx := chi.NewMux(), chi.NewMux()
> 
> rootMx.Use(httplog.RequestLogger(logger))
> apiMx.Use(middleware.JWT)
> apiMx.NotFound(jsonNotFound)
> rootMx.Mount("/api", apiMx)
> 
> cfg := huma.DefaultConfig("My API", "1.0.0")
> 
> cfg.Prefix = "/api" // missing
> 
> api := humachi.New(apiMx, cfg)
> 
> huma.AutoRegister(api, struct{}{})
> 
> huma.AutoRegister()
> ```

Can you describe what the problem? I can't understand.
```go
package main

import (
	"context"
	"flag"
	"fmt"
	"net/http"
	"time"

	"github.com/go-chi/chi/v5"
	"github.com/go-chi/chi/v5/middleware"
	"github.com/go-chi/jwtauth/v5"
)

var (
	tokenAuth *jwtauth.JWTAuth
	port      = flag.Uint("port", 8080, "")
)

func init() {
	tokenAuth = jwtauth.New("HS256", []byte("secret"), nil)

	// For debugging/example purposes, we generate and print
	// a sample jwt token with claims `user_id:123` here:
	_, tokenString, _ := tokenAuth.Encode(map[string]interface{}{"user_id": 123})
	fmt.Printf("DEBUG: a sample jwt is %s\n\n", tokenString)
}

func main() {
	flag.Parse()

	rootMu := chi.NewMux()
	rootMu.Use(middleware.RealIP)
	rootMu.Use(middleware.RequestID)
	rootMu.Use(middleware.Logger)
	rootMu.Use(middleware.Recoverer)

	rootMu.Route("/api", func(apiMu chi.Router) {
		apiMu.Use(jwtauth.Verifier(tokenAuth))
		apiMu.Use(jwtauth.Authenticator(tokenAuth))
		apiMu.Use(middleware.Timeout(10 * time.Second))

		type TestInput struct{}
		type TestOutput struct{}

		apiMu.NotFound(func(w http.ResponseWriter, r *http.Request) {
			w.WriteHeader(http.StatusNotFound)
		})

		config := huma.DefaultConfig("API", "0.1.0")
		api := humachi.New(apiMu, config)
		huma.Get(api, "/", func(ctx context.Context, _ *TestInput) (*TestOutput, error) {
			return &TestOutput{}, nil
		})
	})

	srv := http.Server{
		Addr:    fmt.Sprintf(":%d", *port),
		Handler: rootMu,
	}

	if err := srv.ListenAndServe(); err != nil {
		panic(err)
	}
}
```

PR #350, opened 2024-04-02
feat: support json.RawMessage

### LINKED ISSUE #348: json.RawMessage support
I would like to have some GeoJSON included as part of my request body that is later on deserialized by another library. I would like to use json.RawMessage as a field type. But the validator of Huma is not allowing me that because it states that the input should be a string. Can support for json.RawMessage be included in this library or can I at least disable the validation on a field if I would like to?

As a work arround I could specify the field type as an interface{} marshal it back to bytes and then unmarshal it with my desired library, that feels a bit inefficient and cumbersome.

```
`type CreateRouteBody struct {
	Name        string           `json:"name" example:"Rondje Kralingseplas" doc:"Name to identify a route with" maxLength:"255"`
	RouteType   string           `json:"route_type" enum:"route_points,line_string" doc:"Indication for the type of route"`
	RoutePoints *json.RawMessage `json:"route_points,omitempty" doc:"Feature Collection that contains all the points."`
	LineString  *string          `json:"line_string,omitempty" doc:""`
}` 
```

```
curl --location 'http://localhost:8888/routes/route' \
--header 'Authorization: <token>' \
--header 'Content-Type: application/json' \
--data '{
    "name": "Kralingseplas232",
    "route_type": "route_points",
    "route_points": {
        "type": "FeatureCollection",
        "features": [{
            "type": "Feature",
            "geometry": {
                "type": "Point",
                "coordinates": [125.6, 10.1]
            },
            "properties": {
                "name": "Dinagat Islands"
            }
        }]
    }
}'
```

```
{
    "$schema": "http://localhost:8888/schemas/ErrorModel.json",
    "status": 422,
    "errors": [
        {
            "message": "expected string",
            "location": "body.route_points",
            "value": {
                "features": [
                    {
                        "geometry": {
                            "coordinates": [
                                125.6,
                                10.1
                            ],
                            "type": "Point"
                        },
                        "properties": {
                            "name": "Dinagat Islands"
                        },
                        "type": "Feature"
                    }
                ],
                "type": "FeatureCollection"
            }
        }
    ]
}
```

danielgtaylor:
This adds support for `json.RawMessage` which generates an empty schema (no validation). This means any input is valid and it's up to the handler code to determine if there are any errors.

Fixes #348.

### DIFF
--- a/docs/docs/features/request-inputs.md
+++ b/docs/docs/features/request-inputs.md
@@ -56,6 +56,19 @@ The special struct field `Body` will be treated as the input request body and ca
 
 `RawBody []byte` can also be used alongside `Body` to provide access to the `[]byte` used to validate & parse `Body`.
 
+### Special Types
+
+The following special types are supported out of the box:
+
+| Type              | Schema                                      | Example                       |
+| ----------------- | ------------------------------------------- | ----------------------------- |
+| `time.Time`       | `{"type": "string", "format": "date-time"}` | `"2020-01-01T12:00:00Z"`      |
+| `url.URL`         | `{"type": "string", "format": "uri"}`       | `"https://example.com"`       |
+| `net.IP`          | `{"type": "string", "format": "ipv4"}`      | `"127.0.0.1"`                 |
+| `json.RawMessage` | `{}`                                        | `["whatever", "you", "want"]` |
+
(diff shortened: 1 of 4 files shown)


sven4all:
Looks good to me, this is the behaviour I would expect.

Is it an idea to add it to the documentation as well?

I am not sure where to include it but it could be here https://github.com/danielgtaylor/huma/blob/main/docs/docs/features/request-validation.md or here https://github.com/danielgtaylor/huma/blob/main/docs/docs/how-to/custom-validation.md.

PR #356, opened 2024-04-07
proposal: add base path to config

Yurovskikh:
Hey, I tried to improve the mechanism for adding a base path. If this draft is fine for humachi and humago I would add it for another routers

Based on this PR and comment https://github.com/danielgtaylor/huma/pull/305#issuecomment-2041420945

### DIFF
--- a/adapters/humachi/humachi.go
+++ b/adapters/humachi/humachi.go
@@ -124,6 +124,9 @@ func NewAdapter(r chi.Router) huma.Adapter {
 
 // New creates a new Huma API using the latest v5.x.x version of Chi.
 func New(r chi.Router, config huma.Config) huma.API {
+	if config.BasePath != "" {
+		r.Mount(config.BasePath, r)
+	}
 	return huma.NewAPI(config, &chiAdapter{router: r})
 }
 
--- a/adapters/humago/humago.go
+++ b/adapters/humago/humago.go
@@ -135,7 +135,7 @@ func New(m Mux, config huma.Config) huma.API {
 	if _, ok := v.(interface{ PathValue(string) string }); !ok {
 		panic("This adapter requires Go 1.22+")
 	}
-	return huma.NewAPI(config, &goAdapter{m, ""})
+	return huma.NewAPI(config, &goAdapter{m, config.BasePath})
 }
 
 // NewWithPrefix creates a new Huma API using an HTTP mux with a URL prefix.
--- a/api.go
+++ b/api.go
@@ -150,6 +150,11 @@ type Config struct {
 	// `Info.Version` fields.
 	*OpenAPI
 
+	// BasePath is the base path is the initial URL segment of the API, and
+	// does not include the host name or any additional segments for paths
+	// or operations. It is shared by all operations in the API.
+	BasePath string
+
 	// OpenAPIPath is the path to the OpenAPI spec without extension. If set
 	// to `/openapi` it will allow clients to get `/openapi.json` or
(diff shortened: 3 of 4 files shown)


danielgtaylor:
@Yurovskikh thanks for the PR! Can you help me understand the need for this change? The existing behavior is described at https://huma.rocks/features/bring-your-own-router/#route-groups-base-urls which seems to cover the various use-cases you might need. Can you help me understand what's confusing about it?

This PR looks okay, but I'd like to:

1. Avoid adding things to the config that are already possible via other means (e.g. `config.Servers = ...`)
2. Prevent ambiguity around whether a base path feature would apply to the router at API creation time and how that might interact with the servers slice.

Mainly trying to understand why we need this.

Yurovskikh:
Yes, Huma's documentation describes this use case, but I think explicit is better than implicit. 
Setting the base path from the list of servers, as well as adding it to the route group in Huma, seems to me a less convenient way than setting it once in the config via the base path

Also, it now works differently for different routes: for some routes I should use groups, but for the default http router I have to add them to the constructor.
https://github.com/danielgtaylor/huma/blob/main/adapters/humago/humago.go#L151
This is not a problem, but it may be more consistent for any routers

Yurovskikh:
Yes, this is mainly about API routes, because we declare the API using huma and the router itself just acts as a provider of this scheme.
That’s why I want to reduce work with the router to a minimum
Maybe I'm just misunderstanding the idea. I just saw the same misunderstanding in the issue and tried to improve it

cardinalby:
Sorry for spamming around, wanted to highlight that it's just another example where more general approach I'm suggesting in  https://github.com/danielgtaylor/huma/issues/489 (having already ready and tested code) can help to solve this and similar issues with one [simple conception](https://github.com/cardinalby/hureg/blob/master/docs/reg_middlewares.md) instead of multiple local changes trying to cover all the cases

Just define a group with [base path](https://github.com/cardinalby/hureg/blob/master/docs/base_path.md#-addbasepath).
```go
v1gr := api.            // all operations registered with v1gr will have:
	AddBasePath("/v1").                            // - "/v1" base path
```

You can still use `api` instance that will not add base path to the operations.

[Or even](https://github.com/cardinalby/hureg/blob/master/docs/base_path.md#-addmultibasepaths)

```go
// all operations registered with multiGr will be actually registered twice: with v1 and v2 prefix,
// operationIDs and summaries will be regenerated to correspond the paths to avoid conflicts
multiGr := api.WithMultiBasePaths(nil, "/v1", "/v2")   
```

[Middlewares](https://github.com/cardinalby/hureg/blob/master/pkg/huma/op_handler/middlewares.go), [transformers](https://github.com/cardinalby/hureg/blob/master/docs/transformers.md), [tags](https://github.com/cardinalby/hureg/blob/master/pkg/huma/op_handler/add_tags.go), [security entries](https://github.com/cardinalby/hureg/blob/master/pkg/huma/op_handler/add_security.go), ... etc can be added for a group, and all the operations registered with this group will have them.

Regarding OpenAPI endpoints, they [can be defined manually](https://github.com/cardinalby/hureg/blob/master/docs/openapi_endpoints.md) in the same manner as other operations with possibility to add auth middlewares to them

danielgtaylor:
This should now be resolved with [v2.30.0](https://github.com/danielgtaylor/huma/releases/tag/v2.30.0) which includes much of the functionality from @cardinalby's excellend hureg library. Thanks!

Issue #302, opened 2024-03-12
Access to all headers like `request.Header`


victoraugustolls:
I was developing an OpenTelemetry Tracer middleware for `Huma`, but when I tried propagating the headers, I noticed that I couldn't access all the headers at once to pass to the propagator. I ended up developing a native `chi` middleware instead.

The final result is:

```go
// TracingMiddleware extracts the headers from the request and starts the span.
func TracingMiddleware() func(next http.Handler) http.Handler {
	return func(next http.Handler) http.Handler {
		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			ctx := otel.GetTextMapPropagator().Extract(r.Context(), propagation.HeaderCarrier(r.Header))
			ctx, span := StartSpan(ctx, r.Method)
			r = r.WithContext(ctx)

			defer span.End()

			next.ServeHTTP(w, r)

			span.SetName(fmt.Sprintf("%s %s", r.Method, chi.RouteContext(ctx).RoutePattern()))
		})
	}
}
```

I could implement a custom OTel Propagator for `Huma` to have a router-agnostic middleware, but I wanted your opinion here. My final goal is to provide ready-to-use router-agnostic middleware for `Huma` instrumentation. Do you think `Huma` should give access to the headers like the example above, or should I create a custom propagator for `Huma`? I'm okay with either solution.

victoraugustolls:
That's perfect!  So I will create a custom propagator for Otel. When I finish my middlewares for OpenTelemetry, would you mind if I contributed them to the [official contribution repository](https://github.com/open-telemetry/opentelemetry-go-contrib)? If you prefer, I can put them somewhere else, maybe even in the `Huma` repository.

victoraugustolls:
@danielgtaylor, when writing the agnostic middleware, I noticed that the user could not access the response code after calling `next(ctx)` in the middleware. Is this by design? Or could we include access to the status code?

danielgtaylor:
@victoraugustolls feel free to submit it to the contrib repo, that would be great! As for accessing the response code, you will want to wrap the context with your own implementation that stores the status, something like:

```go
type MyContext struct {
  huma.Context
  status int
}

func (c *MyContext) SetStatus(code int) {
  c.status = code
  c.Context.SetStatus(code)
}
```

Then use that to wrap the incoming context, call `next(wrapped)` and then access `wrapped.status` when submitting metrics.

victoraugustolls:
[...]

PR #369, opened 2024-04-10
Add the ability to get ctx.Status()

ssoroka-tc:
For middlewares that needs to log the status result of a request, or needs to commit/rollback transactions based on the success of the handler, it'd be handy to be able to query the response status. 

In a [recent comment](https://github.com/danielgtaylor/huma/issues/302#issuecomment-1993236171) you mentioned another solution, 
```go
type MyContext struct {
  huma.Context
  status int
}

func (c *MyContext) SetStatus(code int) {
  c.status = code
  c.Context.SetStatus(code)
}
```

Except this doesn't work because `huma.Context` conflicts with the method named `Context()` in the interface definition of `huma.Context`. It gives the error: `cannot use myCtx (variable of type *MyContext) as huma.Context value in argument to next: *MyContext does not implement huma.Context (*MyContext.Context is a field, not a method)`. Perhaps type aliasing or assigning huma.Context to a new type might be a work-around, but we can make this easy for all Huma users.

This PR adds `Status()` to the interface and updates all the adapter contexts.

### DIFF
--- a/adapters/humabunrouter/humabunrouter.go
+++ b/adapters/humabunrouter/humabunrouter.go
@@ -20,10 +20,14 @@ import (
 var MultipartMaxMemory int64 = 8 * 1024
 
 type bunContext struct {
-	op *huma.Operation
-	r  bunrouter.Request
-	w  http.ResponseWriter
+	op     *huma.Operation
+	r      bunrouter.Request
+	w      http.ResponseWriter
+	status int
 }
+
+// check that bunContext implements huma.Context
+var _ huma.Context = &bunContext{}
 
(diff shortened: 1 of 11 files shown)


danielgtaylor:
@ssoroka-tc the quick fix is something like this:

```go
type humaContext huma.Context

type MyContext struct {
  humaContext
  status int
}

func (c *MyContext) SetStatus(code int) {
  c.status = code
  c.humaContext.SetStatus(code)
}
```

Anyway, I agree this has confused enough users that storing an extra int is probably worthwhile, so thanks for the PR!

ssoroka-tc:
Thanks for Huma :D

PR #372, opened 2024-04-11
fix: make cbor support optional

danielgtaylor:
This PR makes the CBOR format optional in Huma's default config, removing the hard dependency and providing more flexibility in how default formats are set up. Technically this is not a breaking change, just a change in how defaults work, however existing users who wish to continue using CBOR must add a new import. If you set up your own config you are unaffected.

- Moves CBOR support into `github.com/danielgtaylor/huma/v2/formats/cbor`
- Creates a `huma.DefaultFormats` map which packages can modify. This is used by `huma.DefaultConfig(...)`.
- Uses the `init()` function to modify the default formats map when the `cbor` package is imported. This is similar to how the standard library works and described in [Effective Go](https://go.dev/doc/effective_go#blank_import), e.g. see [net/http/pprof](https://pkg.go.dev/net/http/pprof) or the `crypto` or `sql` stdlib packages.
- Updates docs & examples

After this change, using Huma to build an API supporting JSON with the Go 1.22 `http.ServeMux` has **zero** hard dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Features include:
 - JSON Errors using [RFC9457](https://datatracker.ietf.org/doc/html/rfc9457) and `application/problem+json` by default (but can be changed)
 - Per-operation request size limits with sane defaults
 - [Content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation) between server and client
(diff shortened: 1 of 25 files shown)


victoraugustolls (approved):
Really liked the approach, LGTM!

x-user:
LGTM.

PR #382, opened 2024-04-14
feat: add per-operation router-agnostic middleware

### LINKED ISSUE #378: Register routes functionality with middlewares
Just discussing. I had a little struggle adding one middleware to different handlers that have different input and output structs. I managed to do it, but I don't like all the wrapping and interfaces I needed to add to make it possible.

Wouldn't it be kinda cool to have a function that works like 'register', but it chains passed middlewares, including their own input, leading to a handler with their input and output?

If something like that already exists, I'm sorry!

danielgtaylor:
This PR introduces a small change to add per-operation router-agnostic middleware using the existing `huma.Operation` struct configured at operation registration time. It can be used like this, which also facilitates wrapping/composition utility functions:

```go
	Method: http.MethodGet,
	Path:   "/demo",
	Middlewares: huma.Middlewares{
		func(ctx huma.Context, next func(huma.Context)) {
			// TODO: implement middleware here...
			next(ctx)
		},
	},
}, func(ctx context.Context, input *struct{}) (*struct{}, error) {
	// TODO: implement handler here...
	return nil, nil
})
```

Let me know what you think. Fixes #378.

### DIFF
--- a/api.go
+++ b/api.go
@@ -224,7 +224,10 @@ type API interface {
 	// the next Middleware.
 	UseMiddleware(middlewares ...func(ctx Context, next func(Context)))
 
-	// Middlewares returns a slice of middleware handler functions.
+	// Middlewares returns a slice of middleware handler functions that will be
+	// run for all operations. Middleware are run in the order they are added.
+	// See also `huma.Operation{}.Middlewares` for adding operation-specific
+	// middleware at operation registration time.
 	Middlewares() Middlewares
 }
 
--- a/docs/docs/features/middleware.md
+++ b/docs/docs/features/middleware.md
@@ -151,6 +151,35 @@ func MyMiddleware(ctx huma.Context, next func(ctx huma.Context)) {
 
     The [`huma.ErrorDetail`](https://pkg.go.dev/github.com/danielgtaylor/huma/v2#ErrorDetail) struct can be used to provide more information about the error, such as the location of the error and the value which was seen.
 
+### Operations
+
+You can also add router-agnostic middleware to individual operations by setting the [`huma.Operation.Middlewares`](https://pkg.go.dev/github.com/danielgtaylor/huma/v2#Operation) field. This middleware will run after the router-specific middleware and before the operation handler.
+
+```go title="code.go"
+func MyMiddleware(ctx huma.Context, next func(huma.Context)) {
+	// Call the next middleware in the chain. This eventually calls the
+	// operation handler as well.
+	next(ctx)
+}
+
+func main() {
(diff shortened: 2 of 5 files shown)


aleksicmarija:
This solves the problem I had, and I think it's good. I'm just curious: would it be beneficial to have inputs and outputs in middlewares, essentially functioning like handlers now? Under the hood, it would resemble chained handlers, but we could use that middleware with multiple handlers.

quadgod:
Thanks! It’s solves my problem too.

ssoroka (approved):

quadgod:
Just my thoughts...

Huma have a [resolvers](https://huma.rocks/features/request-resolvers/). Basically resolver it's just a part of the model, and if you want inject some `stateful` (some logic which interacts with database or some other external provider) logic - it will looks dirty. With middleware i can do it with better way now. And resolvers feature looks useless now & it's adding a complexity to dto models.

victoraugustolls (approved):
LGTM!

danielgtaylor:
@aleksicmarija, I was going to say [request resolvers](https://huma.rocks/features/request-resolvers/) may be what you are looking for. They operate on the input model, can set whatever structured data you need, and can be shared between operations. They run after the middleware but before the handler function.

@quadgod resolvers mostly exist as a way to provide additional validation that cannot be easily modeled via JSON Schema, or a way to translate or modify the input before processing. For example, making sure one field of the input is a multiple of the value in another field (useful for image/video processing) or making sure an argument is title cased while allowing lowercase inputs.

If we want to revisit middleware with typed/validated inputs we can, but I will merge & release this PR so we don't block on that discussion. Feel free to come up with a proposal and post in https://github.com/danielgtaylor/huma/discussions.

aleksicmarija:
> @aleksicmarija, I was going to say [request resolvers](https://huma.rocks/features/request-resolvers/) may be what you are looking for. They operate on the input model, can set whatever structured data you need, and can be shared between operations. They run after the middleware but before the handler function.
> 
> 
> 
> @quadgod resolvers mostly exist as a way to provide additional validation that cannot be easily modeled via JSON Schema, or a way to translate or modify the input before processing. For example, making sure one field of the input is a multiple of the value in another field (useful for image/video processing) or making sure an argument is title cased while allowing lowercase inputs.
> 
> 
> 
> If we want to revisit middleware with typed/validated inputs we can, but I will merge & release this PR so we don't block on that discussion. Feel free to come up with a proposal and post in https://github.com/danielgtaylor/huma/discussions.

Totally missed it, thank you!

Issue #395, opened 2024-04-19
Add support for netip.Addr


wolveix:
Hey!

I see from the following link that `net.IP` is supported as a field type: https://huma.rocks/features/request-inputs/?h=query#special-types

Could we add support for its successor, `netip.Addr`? 

Thanks!

danielgtaylor:
Implemented in #396

PR #396, opened 2024-04-19
Support `netip.Addr` as a special type

wolveix:
Hey! I realised after opening my issue on this (#395), that this should be a pretty trivial issue to resolve. 

Please let me know if I've missed anything with this PR, and thanks again for all that you do!


danielgtaylor (approved):
Thank you :+1:

PR #408, opened 2024-04-25
Use plain http in $schema for 127.0.0.1

davidolrik:
Small quick fix for http schema when using 127.0.0.1


danielgtaylor (approved):
Thank you! :+1:

PR #411, opened 2024-04-26
fix: Add support for Percent-Encoding RFC 3986

chilic:
**Problem**:
The current implementation does not allow to use of encoded named parameters like:
`/path/:param` - `/path-params/a%3A%2F%2Fb%2Fc`

Which is fully legal by [RFC 3986](http://www.ietf.org/rfc/rfc3986.txt)

Solution:
Use [EscapedPath()](https://pkg.go.dev/net/url#URL.EscapedPath) instead of direct access to a `u.Path`

> Note that the Path field is stored in decoded form: /%47%6f%2f becomes /Go/. A consequence is that it is impossible to tell which slashes in the Path were slashes in the raw URL and which were %2f. This distinction is rarely important, but when it is, the code should use the [URL.EscapedPath](https://pkg.go.dev/net/url#URL.EscapedPath) method, which preserves the original encoding of Path.

Details:
https://pkg.go.dev/net/url#URL

PR into the original package:
https://github.com/alexedwards/flow/pull/11

### DIFF
--- a/adapters/humaflow/flow/flow.go
+++ b/adapters/humaflow/flow/flow.go
@@ -165,7 +165,7 @@ func (m *Mux) Group(fn func(*Mux)) {
 
 // ServeHTTP makes the router implement the http.Handler interface.
 func (m *Mux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	urlSegments := strings.Split(r.URL.Path, "/")
+	urlSegments := strings.Split(r.URL.EscapedPath(), "/")
 	allowedMethods := []string{}
 
 	for _, route := range *m.routes {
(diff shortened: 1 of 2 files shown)


danielgtaylor:
@chilic I would assume most users want the decoded param value. What's the use-case for the raw value?

Unless I'm missing something, Go 1.22 behaves the same way (automatically decoding the values), see https://go.dev/play/p/rlvkKwpgs_I for example.

danielgtaylor:
@chilic so sorry I misunderstood what this was about! You are right, the routing is incorrect! However, when getting a path param value it *should* be decoded rather than in its raw encoded state, so I've added a commit for that to ensure param values are decoded before storing them. Let me know what you think and I can merge it in if everything looks okay!

chilic:
I agree with you. Better to keep the behavior the same as the `HTTP` package. Thank you!

Issue #412, opened 2024-04-26
Schema(r huma.Registry) *huma.Schema not working on Pointer


lennycampino:
I want to use the https://github.com/shopspring/decimal library for handling decimals and huma will detect decimal.Decimal as object but it is a string. 

So I implemented the following code: 

```
type Decimal decimal.Decimal

func (d *Decimal) Schema(r huma.Registry) *huma.Schema {
	return &huma.Schema{Type: huma.TypeString}
}

type Request struct {
Amount Decimal `json:"amount"`
}
```
This will work but when Amount becomes a Pointer 
```
type Request struct {
Amount *Decimal `json:"amount"`
}
```

It will not work anymore and Huma will treat the field amount as object again. Could you please help me with that?

PR #413, opened 2024-04-26
fix: check if struct implements interface SchemaProvider after deref

lennycampino:
Problem:
Currently the Schema(r Registry) *Schema on types that implement a custom schema will only be called if the field is not a pointer. 

Solution:
In the function SchemaFromType we first deref the type so that we always compare a non pointer type against the interface. 

fixes: #412


danielgtaylor:
@lennycampino nice fix, thank you! :+1:

Issue #322, opened 2024-03-16
Is there a feature to control how OpenAPI model names are generated ?


lsdch:
After going through the documentation, I now have an urge to make the switch from Gin+Swaggo to Huma. Thanks for the work on this nice piece of tech :pray:

I understand from the [first API](https://huma.rocks/tutorial/your-first-api/#operation) example in the docs that OpenAPI model definitions generated from response input/output body are generated as `{.TypeName}Body`. 
Is there any way to configure this behavior either globally or per response type (e.g. using struct tags) ?

Maybe something like : 
```go
type GreetingOutput struct {
	Body struct {
		Message string `json:"message" example:"Hello, world!" doc:"Greeting message"`
	} `model_name:"Greeting"` // name in OpenAPI spec
}
```

danielgtaylor:
@lsdch it's not possible to do it per response using struct tags (though I think that's a great idea and we could add it), but you can set it globally in the registry. Huma ships with a default naming function:

https://pkg.go.dev/github.com/danielgtaylor/huma/v2#DefaultSchemaNamer

The schema namer functions take a type and a name hint, and return a generated name given those inputs. That in turn is used by the defaults like this:

https://github.com/danielgtaylor/huma/blob/main/defaults.go#L69

You can override it by creating your own config or overwriting the value in the default config, for example:

```go
func MyCustomSchemaNamer(t reflect.Type, hint string) string {
  return "..."
}

config := huma.DefaultConfig("My API", "1.0.0")
config.Components.Schemas = huma.NewMapRegistry("#/components/schemas", MyCustomSchemaNamer)
```

lsdch:
Lovely. Definitely going to switch from Gin to Huma. I might submit a PR to support defining the model name using struct tags once I get familiar with the lib, but that might take a while.

PR #416, opened 2024-04-26
Allow providing name hints using tags for anonymous body structs

lsdch:
Small change introducing the possibility to provide name hints for request/response body through struct tags. Previously discussed in #322

Example: 
```go
type EndpointInput struct {
  Body struct {
    SomeData string `json:"some_data"`
  } `name-hint:"SomeName"`
}
```

Because it defines a name hint, it only applies to anonymous types. We could go further and allow overriding the name altogether but that would involve changing the schema namer interface (`func (t reflect.Type, hint string) string`) to include one more parameter, which might break backward compatibility.

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -582,8 +582,11 @@ func Register[I, O any](api API, op Operation, handler func(context.Context, *I)
 			if c := f.Tag.Get("contentType"); c != "" {
 				contentType = c
 			}
-
-			s := SchemaFromField(registry, f, getHint(inputType, f.Name, op.OperationID+"Request"))
+			hint := getHint(inputType, f.Name, op.OperationID+"Request")
+			if nameHint := f.Tag.Get("nameHint"); nameHint != "" {
+				hint = nameHint
+			}
+			s := SchemaFromField(registry, f, hint)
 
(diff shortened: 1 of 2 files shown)


huma.go
```diff
@@ -582,8 +582,11 @@ func Register[I, O any](api API, op Operation, handler func(context.Context, *I)
 			if c := f.Tag.Get("contentType"); c != "" {
 				contentType = c
 			}
-
-			s := SchemaFromField(registry, f, getHint(inputType, f.Name, op.OperationID+"Request"))
+			hint := getHint(inputType, f.Name, op.OperationID+"Request")
+			if nameHint := f.Tag.Get("name-hint"); nameHint != "" {
```
danielgtaylor, line 586: One small nit: our existing tags (e.g. see https://huma.rocks/features/request-validation/#validation-tags) use `lowerCamelCasing` so we should probably use `nameHint` for consistency.

danielgtaylor (commented):
This is awesome, thank you!

huma.go
lsdch: Alright !  Changed it to `nameHint`

danielgtaylor (approved):

PR #415, opened 2024-04-26
Feature: handling files from `multipart/form-data` request

lsdch:
This PR aims to improve the handling of `multipart/form-data` requests by: 

- 3570bf1 allowing users to document the request body at the level of operations, which is currently not possible since it is overwritten by huma when `RawBody` is `multipart.FormData`. The [generated schema](https://github.com/danielgtaylor/huma/blob/ed2c3cd81698cb5728a49c7cfa267f23c392f644/huma.go#L633) is more of an example and does not reflect the actual form data structure, which should be specified by the user.
- d6eb219 implementing `MultipartFormFiles`, which can be used to type `RawBody` in request inputs and provides logic for decoding and validating files from a `multipart/form-data` request as `multipart.File` values, as well as generating the associated schema in the spec. Although this could be extended to support data types other than files, it would require significant refactoring to reuse some of the logic from body/query/path/header parameters decoding, and I was not confident enough to engage into it.
- 01ca953 implementing mimetype validation for submitted files

There is currently no automated tests for this PR, I would like to first get some feedback before spending more time on this (and still have to get familiar with `humatest`).

Example usage: 
```go
type FileData struct {
        // This is an example, any number of `multipart.File` fields can be defined.
        // Nested structs are not supported.
	SomeFile multipart.File   `form-data:"some_file" content-type:"image/png" required:"true"`
	SeveralFiles []multipart.File `form-data:"several_files" content-type:"image/png,image/jpeg" required:"true"`
}

type FileHandlerInput struct {
	RawBody huma.MultipartFormFiles[FileData]
}

func FileHandler(ctx context.Context, input *FileHandlerInput) (*struct{}, error) {
  fileData := input.RawBody.Data()
  DoSomeThingWith(fileData.SomeFile)
  OrSomethingElseWith(fileData.SeveralFiles)
}

huma.Register(api,
  huma.Operation{
	  Path:             "/handle-files",
	  Method:        http.MethodPost,
	  OperationID:     "Handle files",
  }, FileHandler)
```

When using `MultipartFormFiles[T]`, only the fields in `T` that implement `multipart.File` or `[]multipart.File` are considered when generating the spec or resolving a request. The raw form data can be accessed at `MultipartFormFiles.Form` which is a `multipart.Form`.

`form-data` tag specifies the lookup key to retrieve the encoded file in the `multipart.Form`. In its absence, the name of the field is used instead. 
[...]

PR #422, opened 2024-04-29
Add contentMediaType field to generated schema when format is binary

lsdch:
This PR adds the `contentMediaType` property to the generated schema spec for binary field. This is mainly for the purpose of compatibility with API documentation tools such as Stoplight Elements (see #397).

Note that #415 sets `contentType` property in the schema, but that is not sufficient for Stoplight to show the file upload UI component, event though the [OpenAPI spec](https://spec.openapis.org/oas/v3.1.0#considerations-for-file-uploads) says `contentMediaType` is not required when `contentType` is defined. 
Although that may rather be an issue in Stoplight, in my understanding it does not hurt to add this redundancy to Huma.

Thus, this PR can be considered independently from #415. I hope I'm not mistaken about what the OpenAPI specs say, let me know if you think this rather belongs to Stoplight issues.

Example schema: 

```json
{
  "requestBody": {
	"content": {
		"multipart/form-data": {
			"schema": {
				"properties": {
					"filename": {
						"contentMediaType": "application/octet-stream",
						"description": "filename of the file being uploaded",
						"format": "binary",
						"type": "string"
					},
					"name": {
						"description": "general purpose name for multipart form value",
						"type": "string"
					}
				},
				"type": "object"
			}
		}
	},
	"required": true
}
```

Sample UI screenshot: 
image

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -154,12 +154,17 @@ func (s *Schema) MarshalJSON() ([]byte, error) {
 	if s.Nullable {
 		typ = []string{s.Type, "null"}
 	}
+	var contentMediaType string
+	if s.Format == "binary" {
+		contentMediaType = "application/octet-stream"
+	}
 	return marshalJSON([]jsonFieldInfo{
 		{"type", typ, omitEmpty},
 		{"title", s.Title, omitEmpty},
 		{"description", s.Description, omitEmpty},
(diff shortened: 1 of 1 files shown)


danielgtaylor (approved):
Thanks!

PR #431, opened 2024-05-04
panic with more descriptive messages

ssoroka:
Ran into this problem where a dev tried to use *int for a response, and the panic wasn't clear about what to fix. This tries to point the dev in the right direction.


danielgtaylor (approved):
Thanks! :+1:

PR #434, opened 2024-05-10
feat: convenience methods with operationHandlers

fourcels:


danielgtaylor (commented):
@fourcels thanks for the PR! I like this approach and we can add on to it later as needed with additional functionality.

PR #438, opened 2024-05-11
Feature add converter for user types

### LINKED ISSUE #433: Allow custom query param types.
The idea is to have something like 
```go
type Nullable[T any] struct {
	Null  bool
	Value T
}

func (o Nullable[T]) Schema(r huma.Registry) *huma.Schema {
	return r.Schema(reflect.TypeOf(o.Value), true, "")
}

type CustomQueryParamInput struct {
	Query Nullable[int] `query:"query"`
}
```
And allow the user to implement the field conversion.

Using the SchemaProvider interface it can be set the query param type, but the current implementation does not provide any interface for the conversion part.

The fallback is the TextUnmarshaler interface, but this interface is not enough because then will fail on the validation phase (always set the value as a string).

So my proposal is to add an interface that allow the user to implement this conversion.. something like
```go
type InputConverter interface {
	HumaInputConvert([]byte) (any, error)
}
```

the relevant code is on the huma.go file, starting at line 825.

ross96D:
fix #433

### DIFF
--- a/api.go
+++ b/api.go
@@ -231,6 +231,10 @@ type API interface {
 	Middlewares() Middlewares
 }
 
+type InputParamConverter interface {
+	HumaInputParamConvert([]byte) (any, error)
+}
+
 // Format represents a request / response format. It is used to marshal and
 // unmarshal data.
 type Format struct {
--- a/huma.go
+++ b/huma.go
@@ -1388,6 +1388,37 @@ func getValue(f reflect.Value, p *paramFieldInfo, value string, res *ValidateRes
 		return v, true
 	default:
 		if f.Type().Kind() == reflect.Slice {
+			typ := f.Type().Elem()
+			interfaceType := reflect.TypeFor[InputParamConverter]()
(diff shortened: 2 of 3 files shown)


danielgtaylor:
@ross96D thanks for the PR. I was looking into this one last night and wanted to give a brief update. I think you correctly identified a problem with custom param validation. I'm not sure a second method to convert from `[]byte` is the best approach, so I'm thinking about what would be best for this feature. I thought about a custom `Validate` method but it wouldn't be used for body validation as that happens before the inputs are parsed into the struct values. 🤔 ... We could just return a value via some method so it can be used for validation. I'll try to propose something for this soon.

danielgtaylor:
I think this should now be handled in a fast way without additional memory allocations by https://github.com/danielgtaylor/huma/releases/tag/v2.29.0 param enhancements like the `ParamWrapper` interface to return the `reflect.Value` to set. Please let me know if that doesn't resolve things!

PR #440, opened 2024-05-16
fix: unmarshaling of request body when using struct embedding

lsdch:
Stumbled on this bug while trying to refactor some handler inputs. When the request `Body` is defined within an embedded struct, unmarshaling of the request body does not work as expected, because the field is retrieved by index which points to the embedded struct, not its `Body` content. 

Small example to illustrate the issue : 
```go
type BodyContainer struct {
  Body struct {
    Name string `json:"name"`
 }
}

type HandlerInput struct {
  BodyContainer
}

func Handler(ctx context.Context, input HandlerInput) (*struct{}, error) {
  fmt.Printf("%+v", input.Body) // Name is empty string even when it is not in the JSON payload
  return nil, nil
}
```

## Summary by CodeRabbit

- **Refactor**
  - Improved internal handling of input body validation and conversion.

- **Tests**
  - Enhanced test cases for request body unmarshalling with a new `BodyContainer` struct.

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -569,9 +569,11 @@ func Register[I, O any](api API, op Operation, handler func(context.Context, *I)
 		panic("input must be a struct")
 	}
 	inputParams := findParams(registry, &op, inputType)
-	inputBodyIndex := -1
+	inputBodyIndex := make([]int, 0)
+	hasInputBody := false
 	if f, ok := inputType.FieldByName("Body"); ok {
-		inputBodyIndex = f.Index[0]
+		hasInputBody = true
+		inputBodyIndex = f.Index
 		if op.RequestBody == nil {
 			required := f.Type.Kind() != reflect.Ptr && f.Type.Kind() != reflect.Interface
 			if f.Tag.Get("required") == "true" {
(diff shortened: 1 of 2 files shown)


huma.go
```diff
@@ -1226,7 +1226,9 @@ func Register[I, O any](api API, op Operation, handler func(context.Context, *I)
 						// second time is faster than `mapstructure.Decode` or any of the other
 						// common reflection-based approaches when using real-world medium-sized
 						// JSON payloads with lots of strings.
-						f := v.Field(inputBodyIndex)
+						// Body struct field is retrieved by name to work with embedded structs,
+						// as its existence has already been verified.
+						f := v.FieldByName("Body")
```
danielgtaylor, line 1231: Nice find for this bug! One small nit: based on benchmarks it is much faster to get a field by index, so I think if the body is in an embedded struct we should store a slice of indexes to access it rather than using `FieldByName`. Otherwise I think this PR is great!

lsdch: That makes sense, I patched the PR following your suggestion. I'm confused about why the code coverage check fails tho

danielgtaylor (approved):

PR #441, opened 2024-05-16
feat: allow toggling additional properties from request body struct tag

lsdch:
This is a small QoL improvement that allows toggling the `additionalProperties` attribute on a request body, without having to modify the body struct type. 

Usage : 
```go
type HandlerInput struct {
  Body SomeInputType `additionalProperties:"true"`
}
```

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -587,6 +587,12 @@ func Register[I, O any](api API, op Operation, handler func(context.Context, *I)
 				hint = nameHint
 			}
 			s := SchemaFromField(registry, f, hint)
+			if _, ok = f.Tag.Lookup("additionalProperties"); ok {
+				if s.Ref != "" {
+					s = registry.SchemaFromRef(s.Ref)
+				}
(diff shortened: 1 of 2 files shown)


danielgtaylor:
@lsdch this is a good idea but I didn't add this because of the following scenario:

```go
type MyBody struct {
  Field string `json:"field"`
}

type Response1 struct {
  Body MyBody `additionalProperties:"true"`
}

type Response2 struct {
  Body MyBody `additionalProperties:"false"`
}
```

How would you generate the OpenAPI spec for this? In JSON Schema the `additionalProperties` field is a property of the `object` schema rather than on the individual properties within some object. You _could_ use `allOf` to combine the object definition with `additionalProperties`, but that has pretty undesirable effects for code generated SDKs.

lsdch:
@danielgtaylor ah yes, totally agree thanks for pointing this out. Closing!

PR #462, opened 2024-06-03
fix: type schema description not work

fourcels:
```go
type LockType uint

func (*LockType) Schema(r huma.Registry) *huma.Schema {
	schema := huma.SchemaFromType(r, reflect.TypeOf(0))
	schema.Enum = []any{NFC, Bluetooth}
	schema.Description = "description"
	return schema
}

```


danielgtaylor (approved):
Thanks! :+1:

PR #464, opened 2024-06-04
fix: schema enum validation

fourcels:
```go
type LockType uint

const (
	NFC LockType = iota + 1
	Bluetooth
)

func (*LockType) Schema(r huma.Registry) *huma.Schema {
	schema := huma.SchemaFromType(r, reflect.TypeOf(0))
	schema.Enum = []any{NFC, Bluetooth}
	return schema
}
```

```json
{
  "status": 422,
  "errors": [
    {
      "message": "expected value to be one of \"1, 2\"",
      "location": "body.type",
      "value": 1
    }
  ]
}

```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation logic to ensure correct comparison of enum values.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -487,8 +487,9 @@ func Validate(r Registry, s *Schema, path *PathBuffer, mode ValidateMode, v any,
 
 	if len(s.Enum) > 0 {
 		found := false
+		vs := fmt.Sprint(v)
 		for _, e := range s.Enum {
-			if e == v {
+			if fmt.Sprint(e) == vs {
 				found = true
 				break
 			}


ross96D:
@danielgtaylor i think this issue relates to #438. A good solution on #438 solves the problem here and avoids the use of a fmt.Sprint which allocates and seems a bit hacky

fourcels:
@codercms have a try
```go
func (*VehicleType) Schema(r huma.Registry) *huma.Schema {
	schema := huma.SchemaFromType(r, reflect.TypeFor[string]()) // always complains that the validation failed
	schema.Enum = []any{string(VehicleTypeCar), string(VehicleTypeTruck), string(VehicleTypeSpecial)}
	return schema
}

```

codercms:
@fourcels it works, thanks :)

PR #469, opened 2024-06-05
Extend TestAPI interface to allow for custom context.Context

coury-clark:
This PR extends the humatest.TestAPI interface to allow for a custom context.Context to be passed into the request. This is useful for applications where the context is used to store / propagate request specific values, such as authentication / authorization information. This is essentially porting a wrapper I had to write internally back into the mainline, it seems universally useful enough to me - let me know if you disagree!

This does introduce a breaking change in the TestAPI interface, which is difficult to avoid given the `humatest.New` function returns the interface. If that isn't acceptable, please let me know!

Alternative naming options are similar to `http` package - ie. GetWithContext. Let me know if you'd prefer that style.

Thanks for huma!

## Summary by CodeRabbit

- **New Features**
  - Introduced context handling in API methods to support more advanced request scenarios.
  - Added new `Ctx` methods for `Do`, `Get`, `Post`, `Put`, `Patch`, and `Delete` to accept a `context.Context` parameter.

- **Documentation**
  - Updated the tutorial to include a new section on using custom `context.Context` in Go tests for scenarios like testing authenticated routes.

### DIFF
--- a/humatest/humatest.go
+++ b/humatest/humatest.go
@@ -236,7 +236,7 @@ func (a *testAPI) PatchCtx(ctx context.Context, path string, args ...any) *httpt
 
 func (a *testAPI) Delete(path string, args ...any) *httptest.ResponseRecorder {
 	a.tb.Helper()
-	return a.Do(http.MethodDelete, path, args...)
+	return a.DeleteCtx(context.Background(), path, args...)
 }
 
 func (a *testAPI) DeleteCtx(ctx context.Context, path string, args ...any) *httptest.ResponseRecorder {
--- a/humatest/humatest.go
+++ b/humatest/humatest.go
@@ -4,6 +4,7 @@ package humatest
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -41,13 +42,16 @@ func NewAdapter() huma.Adapter {
 type TestAPI interface {
 	huma.API
 
-	// Do a request against the API. Args, if provided, should be string headers
+	// DoCtx a request against the API. Args, if provided, should be string headers
 	// like `Content-Type: application/json`, an `io.Reader` for the request
 	// body, or a slice/map/struct which will be serialized to JSON and sent
 	// as the request body. Anything else will panic.
+	DoCtx(ctx context.Context, method, path string, args ...any) *httptest.ResponseRecorder
+
+	// Do wraps [TestAPI.DoCtx] using [context.Background].
 	Do(method, path string, args ...any) *httptest.ResponseRecorder
 
-	// Get performs a GET request against the API. Args, if provided, should be
+	// GetCtx performs a GET request against the API. Args, if provided, should be
 	// string headers like `Content-Type: application/json`, an `io.Reader`
 	// for the request body, or a slice/map/struct which will be serialized to
 	// JSON and sent as the request body. Anything else will panic.
@@ -57,9 +61,12 @@ type TestAPI interface {
 	//
 	// 	// Make a GET request with a custom header.
 	// 	api.Get("/foo", "X-My-Header: my-value")
+	GetCtx(ctx context.Context, path string, args ...any) *httptest.ResponseRecorder
+
+	// Get wraps [TestAPI.GetCtx] using [context.Background].
 	Get(path string, args ...any) *httptest.ResponseRecorder
 
-	// Post performs a POST request against the API. Args, if provided, should be
+	// PostCtx performs a POST request against the API. Args, if provided, should be
 	// string headers like `Content-Type: application/json`, an `io.Reader`
 	// for the request body, or a slice/map/struct which will be serialized to
 	// JSON and sent as the request body. Anything else will panic.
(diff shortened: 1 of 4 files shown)


humatest/humatest.go
```diff
@@ -81,9 +91,12 @@ type TestAPI interface {
 	//
 	// 	// Make a PUT request with a custom header.
 	// 	api.Put("/foo", "X-My-Header: my-value", MyBody{Foo: "bar"})
+	PutCtx(ctx context.Context, path string, args ...any) *httptest.ResponseRecorder
+
+	// Put wraps [TestAPI.PutCtx] using [context.Background].
```
ccoVeille, line 96: For this one and others I think you should address these:

- the comment should be taken from existing code, here your changes lead to have less information in the comment, so in the godoc
- you should add something to existing comment to mention PutCtx should be preferred when there is an existing context

humatest/humatest.go
```diff
@@ -167,27 +192,52 @@ func (a *testAPI) Do(method, path string, args ...any) *httptest.ResponseRecorde
...
+
+func (a *testAPI) PutCtx(ctx context.Context, path string, args ...any) *httptest.ResponseRecorder {
+	a.tb.Helper()
+	return a.DoCtx(ctx, http.MethodPut, path, args...)
 }
 
 func (a *testAPI) Patch(path string, args ...any) *httptest.ResponseRecorder {
 	a.tb.Helper()
-	return a.Do(http.MethodPatch, path, args...)
+	return a.PatchCtx(context.Background(), path, args...)
```
ccoVeille, line 225: Here and for all existing code you update.

I would expect to see the context defined as a variable, with a comment explaining why you need this

I would also expect the code to add a `//nolint: context check // some explanation`

To avoid any linting issue

As you may duplicate this code and comment, maybe creating a function doing that would be a good thing.

humatest/humatest.go
```diff
@@ -69,9 +76,12 @@ type TestAPI interface {
 	//
 	// 	// Make a POST request with a custom header.
 	// 	api.Post("/foo", "X-My-Header: my-value", MyBody{Foo: "bar"})
```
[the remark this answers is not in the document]
coury-clark: Thanks for catching that, I've updated the comments to reflect the new signature.

humatest/humatest.go
coury-clark: Ah, I didn't realize comments inside interfaces don't support links in pkgsite (it does in most editors, for example). I've copied the comment over and edited them to reflect the correct signature.

humatest/humatest.go
coury-clark: I'm not entirely sure I follow this feedback, let me know if I'm misunderstanding the comment. Passing context.Background is pretty common as an argument (for example, [from http](https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/net/http/request.go;l=856), or from [this repo](https://github.com/coury-clark/huma/blob/e55b175a0293854df7e43bafe48679667e33322f/examples/sse/main.go#L183)). I don't see any context check linters configured in the repo, and don't see this lint exclusion elsewhere in the repo. 

The linter doesn't seem to mind for me either:
``` 
➜  huma git:(tests-with-ctx) ✗ golangci-lint run
registry.go:140:33: directive `//nolint:musttag` is unused for linter "musttag" (nolintlint)
        return json.Marshal(r.schemas) //nolint:musttag
                                       ^
validate_test.go:1292:67: directive `//nolint: musttag` is unused for linter "musttag" (nolintlint)
                                schemaJSON, _ := json.MarshalIndent(registry.Map(), "", "  ") //nolint: musttag

```

I think this (context.Background()) is a pretty common thing to see, let me know.

ccoVeille: You can ignore this comment.

I'm a linter fanatic. I tend to write everything as if it would be linted with all option on.

humatest/humatest.go
ccoVeille: I think you missed or ignored my second comment.

The fact you could use in each old code there is a version with a context that should be preferred

ccoVeille (approved):
One small remark, nothing really important, so I'm OK 👍

humatest/humatest.go
coury-clark: Ah, sorry it seemed I missed or ignored this. Let me clarify!

The comments do make some case for the scenarios you'd use it, perhaps implicitly
> Get performs a GET request against the API using [context.Background] in the
	// [http.Request]
	
While it's true this is an implicit instruction, I'd expect anyone familiar with Go to understand when they do or do not need to provide their own context.

To address preference, I don't necessarily think from the perspective of the library there is a preference - this is simply a utility after all. Similarly, `http` has no preference if users use `NewRequest` or `NewRequestWithContext`, it's just an option. I'd expect anyone using this library to be able to make their own choices about whether or not they need to provide a context. That said, I didn't update the website documentation (which in practice, is where I get most of the examples from), so please look at https://github.com/danielgtaylor/huma/pull/469/commits/da24dacc085ae5188f67114a6345dd85d65db609.

ccoVeille:
LGTM 👍

danielgtaylor (approved):
@coury-clark this looks great, thank you! :+1:

FYI you can already do something similar by making your own `http.Request` and then using `api.Adapter().ServeHTTP(w, r)` but this is much simpler!

PR #472, opened 2024-06-06
Update min Go version for tutorial

DanRoscigno:
With Go < 1.22 I get this error when running the tutorial:

```
../../go/pkg/mod/github.com/danielgtaylor/huma/v2@v2.18.0/formdata.go:173:21:

undefined: slices.Concat
```
I think slices.Concat is new in 1.22. With 1.22.4 the tutorial works fine.

I am a tech writer by day, using Go at night to write tests of the docs that I write. I could be completely wrong about the need for 1.22, but here is my reference: https://blog.carlana.net/post/2024/golang-slices-concat/

## Summary by CodeRabbit

- **Documentation**
  - Updated the required version of Go for Huma from 1.20 to 1.22 in the installation tutorial.

### DIFF
--- a/docs/docs/tutorial/installation.md
+++ b/docs/docs/tutorial/installation.md
@@ -6,7 +6,7 @@ description: Learn how to install Huma and create your first API.
 
 ## Prerequisites
 
-Huma requires [Go 1.20 or newer](https://go.dev/dl/), so install that first. You'll also want some kind of [text editor or IDE](https://code.visualstudio.com/) to write code and a terminal to run commands.
+Huma requires [Go 1.22 or newer](https://go.dev/dl/), so install that first. You'll also want some kind of [text editor or IDE](https://code.visualstudio.com/) to write code and a terminal to run commands.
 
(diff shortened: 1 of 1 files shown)


danielgtaylor:
@DanRoscigno thanks for pointing this out. This is actually a bug as we want to support Go 1.20+. I'll see what to do about this. I know at some point we'll want to move to Go 1.21+ and the `slices` package was introduced in 1.21.

SirWaithaka:
@danielgtaylor in the similar blog post shared by @DanRoscigno, the following is pointed out.

> It’s worth noting that if you only want to concatenate two slices, there’s no need to use slices.Concat. You can use the built in [append](https://pkg.go.dev/builtin#append) function with the ... spread operator instead:
```concat := append(s1, s2...)```

Perhaps this is the best approach to fix the bug, similar to https://github.com/danielgtaylor/huma/blame/4a07c36b06f16d507b5f0e18a8a61fc4ed9e8112/formdata.go#L174

Issue #452, opened 2024-05-24
Support multiple query parameters with the same name


csmarchbanks:
I am looking at migrating to use huma in a project but one thing that is blocking me is support for requests such as `/path?foo=bar&foo=baz`. This is especially useful when a value may include a comma which would cause the current comma based approach to fail.

I am happy to implement this feature, and made a start, however I am not sure the best way to get the multiple values out of `huma.Context`. I could add a new method such as `QueryMultiple(name string) []string` to the interface, but would that be a breaking change for anyone implementing their own adapter? Happy to hear any suggestions. Worst case I suppose we could cast to a second interface and throw a runtime error?

danielgtaylor:
@csmarchbanks I'm open to trying to support this. I didn't originally implement it because of performance reasons, i.e. you can avoid all dynamic allocations in the request handling hot path by only having one value per query param name. I think returning `[]string` will require allocations, but I'm open to ideas. The benchmarks in `adapters/humachi` can be helpful to see how your changes impact memory and garbage collection.

turtledev1:
I'm trying to support the same thing. Currently it is supported by OpenAPI with the parameters "explode: true" and "style: form" (https://spec.openapis.org/oas/v3.1.0#parameter-object). I tried to add these annotations but the resulting json always have the explode property to true, until I stumbled on this thread.

ARolek:
I'm also encountering this block. I'm happy to help with code review / testing / etc.

csmarchbanks:
I finally had some time to implement this and have a PR up at https://github.com/danielgtaylor/huma/pull/498, let me know if you run into any issues with it!

PR #498, opened 2024-07-11
Implement exploded query parameters

csmarchbanks:
Allow query parameters such as /foo?bar=baz&bar=qux rather than requiring multiple values to be comma separated. This has worse performance than comma seprated so is not enabled by default and requires adding the `explode` tag to the query parameter.

Closes: #452 

## Summary by CodeRabbit

- **New Features**
	- Introduced support for specifying the same query parameter multiple times using the `explode` tag in URI query strings, offering an alternative to comma-separated lists for multiple values.
- **Documentation**
	- Added documentation for the new query parameter `explode` feature.

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -102,6 +102,7 @@ type paramFieldInfo struct {
 	Required   bool
 	Default    string
 	TimeFormat string
+	Explode    bool
 	Schema     *Schema
 }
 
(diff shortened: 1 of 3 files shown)


huma.go
```diff
@@ -136,11 +137,14 @@ func findParams(registry Registry, op *Operation, t reflect.Type) *findResult[*p
...
 			pfi.Loc = "query"
-			name = q
+			split := strings.Split(q, ",")
+			name = split[0]
 			// If `in` is `query` then `explode` defaults to true. Parsing is *much*
-			// easier if we use comma-separated values, so we disable explode.
-			nope := false
-			explode = &nope
+			// easier if we use comma-separated values, so we disable explode by default.
+			if slicesContains(split[1:], "explode") {
```
ARolek, line 144: Consider using the built in slices package: https://pkg.go.dev/slices#Contains

ARolek (commented):
Looks really clean in there. I left a minor suggestion inline. I was trying to come up with alternative names for `Explode`, but didn't come up with anything better. The name works, but could maybe be more specific. Thanks for tackling this issue. It will help me get through a blocker.

csmarchbanks:
>  I was trying to come up with alternative names for Explode, but didn't come up with anything better. The name works, but could maybe be more specific

Thinking about this some more over the weekend, maybe `repeated` would be better?

ARolek:
> Thinking about this some more over the weekend, maybe repeated would be better?

That could work. I know it's a bit verbose, but "Support Multiples" is potentially more specific.

huma.go
danielgtaylor: This is to support Go 1.20 FYI. Once Go 1.23 is released I will probably make the minimum required version 1.21 and we can start using the `slices` package.

danielgtaylor (approved):
LGTM, thanks for making this optional!

PR #520, opened 2024-07-24
feat: make validation error messages customizable

smacker:
Hello!

In the project I'm working on we show validation errors directly to the end user.
In this project we need to return them in specific format defined by the product.

But currently validation error messages are hard-coded in the source code so we would need to either replace validation completely or rely on existing text messages that might change over time.

This PR tries to improve it by allowing:
1. Ability to compare errors safely using variables
2. Customize error messages by re-defining variables during application start
3. Customize formatting function to further customize error message

For example we want to replace existing error:
```
expected string to be RFC 5322 email: mail: missing '@' or angle-addr
```
with something simple like
```
incorrect email address
```

the following code let us do it
```go
func init() {
	// override default error messages
	validation.MsgExpectedRFC5322Email = "incorrect email address"
	validation.ErrorFormatter = func(format string, a ...any) string {
		if format == validation.MsgExpectedRFC5322Email {
			return validation.MsgExpectedRFC5322Email
		}

		return fmt.Sprintf(format, a...)
	}
}
```

I'm not sure if we need unit tests for this and what exactly we need to test.

Please let me know if the idea resonates with you or you have something else in mind. Thank you!

## Summary by CodeRabbit

- **New Features**
  - Introduced a centralized set of validation error messages for improved consistency and clarity across various validation scenarios.
  - Added an `ErrorFormatter` function to enhance error message formatting dynamically.
  
- **Bug Fixes**
  - Enhanced error message handling by replacing hardcoded strings with standardized constants, ensuring uniformity and reducing the risk of inconsistencies.
  
- **Tests**
  - Added a new test for validating custom error formatting, improving test coverage for validation logic.


validate.go
```diff
@@ -166,16 +168,6 @@ func (r *ValidateResult) Add(path *PathBuffer, v any, msg string) {
 	})
 }
 
-// Addf adds an error to the validation result at the given path and with
-// the given value, allowing for fmt.Printf-style formatting.
-func (r *ValidateResult) Addf(path *PathBuffer, v any, format string, args ...any) {
```
danielgtaylor, line 171: Given this is part of the public API and we can't know if others are depending on it, I don't think this can be removed even if we stop using it internally, just FYI.

validation/messages.go
```diff
@@ -0,0 +1,50 @@
...
+	MsgExpectedMinItems                   = "expected array length >= %d"
+	MsgExpectedMaxItems                   = "expected array length <= %d"
+	MsgExpectedMinProperties              = "expected object with at least %d properties"
+	MsgExpectedMaxProperties              = "expected object with at most %d properties"
+	MsgExpectedRequiredProperty           = "expected required property %s to be present"
+	MsgExpectedDependentRequiredProperty  = "expected property %s to be present when %s is present"
+)
+
+// ErrorFormatter is a function that formats an error message
+var ErrorFormatter func(format string, a ...any) string = fmt.Sprintf
```
danielgtaylor, line 50: Random thought: what about making this aware of the request context to enable i18n/l10n scenarios? For example, you could then use a library to return a localized error message based on the request's `Accept-Language` header. As is with your PR you could provide a single langauge, what I'm suggesting here is providing many languages and choosing one based on the request's preferences.

danielgtaylor (commented):
Overall I think this looks great, thanks! One small comment and I want the tests/coverage checks to pass but otherwise this looks mostly ready.

validation/messages.go
smacker: Actually it is the next change I want to submit :) But decided to split the work into smaller pieces.

I noticed you try to avoid strings allocation during request validation. Using ctx would break it because we won't be able to do pre-computation.

The current approach I'm trying is to define 2 interfaces:
```go
// ErrorFormatter is a function that formats an error message
var ErrorFormatter func(format string, a ...any) string = fmt.Sprintf

// ErrorFormatterContext is a function that will be used instead of ErrorFormatter
// to format an error message when defined
var ErrorFormatterContext func(ctx Context, format string, a ...any) string = nil
```

Usage of `ErrorFormatterContext` is opt-in. I didn't do benchmarking yet though. If you don't mind I'll continue with it in the following weeks.

validate.go
smacker: oh. You are right! Sorry about that.

validation/messages.go
```diff
@@ -0,0 +1,45 @@
...
+	MsgExpectedMaxLength                  = "expected length <= %d"
+	MsgExpectedBePattern                  = "expected string to be %s"
+	MsgExpectedMatchPattern               = "expected string to match pattern %s"
+	MsgExpectedMinItems                   = "expected array length >= %d"
+	MsgExpectedMaxItems                   = "expected array length <= %d"
+	MsgExpectedMinProperties              = "expected object with at least %d properties"
+	MsgExpectedMaxProperties              = "expected object with at most %d properties"
+	MsgExpectedRequiredProperty           = "expected required property %s to be present"
+	MsgExpectedDependentRequiredProperty  = "expected property %s to be present when %s is present"
+)
```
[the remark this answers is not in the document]
smacker: the errors messages are used as is to avoid potential breaking changes

validation/messages.go
smacker: I also moved it to the root of huma. It would make code easier to read as we introduce context.

smacker:
> I want the tests/coverage checks to pass but otherwise this looks mostly ready.

I added a test for the new code but I believe it won't improve the report from codecov. 
The few lines that are currently not covered don't change in the PR but because the total number of lines has changed it reports the drop in coverage (I guess?).

I would really appreciate if you could help me to understand how to resolve this issue.
Or I can just add a test to unrelated code, it would bump the coverage a little and probably make codecov happy.

validation/messages.go
danielgtaylor: Good point about the pre-allocation. This probably requires a bit more thought, so I'm happy to get this improvement in first!

danielgtaylor (approved):

Issue #457, opened 2024-05-29
empty array will be send out as null


ScriptType:
i have following type which im returning.
type InstanceArrayResponse struct {
    Body []*InstanceBody
}

if the Body is now an empty array i get null as a response when testing the api.

fmt.Println(instancesResponse)   //&{[]}

so &{[]} is what i`m returning in the handler but in the frontend i receive null :/

When the array is Filled with values its working as expected.

danielgtaylor:
@ScriptType this is how the Golang JSON encoder works:

> Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value.

From https://pkg.go.dev/encoding/json#Marshal

What should the default behavior be? Should it be to send no body, `null`, or `[]` / `{}` when the body is `nil` or empty?

bshtmichielsen:
Hi both I experienced a similar situation and was a little surprised too. I would have expected that if `Body == nil` it outputs `null`, however when Body contains an empty slice, I would have expected `[]` as output.

ScriptType:
> @ScriptType this is how the Golang JSON encoder works:
> 
> > Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value.
> 
> From https://pkg.go.dev/encoding/json#Marshal
> 
> What should the default behavior be? Should it be to send no body, `null`, or `[]` / `{}` when the body is `nil` or empty?

Yep You are right didn't read the docs well enough, my bad.(and im pretty new to go)
Thought it was a Huma thing as i couldn't think of it to be the "default" way.

It still Suprises me to be honest :P.

But yea not a Hma problem so closed

PR #527, opened 2024-07-29
fix: nullable schemas for arrays/slices

lsdch:
Hi, this is a small fix regarding the schema generation for array/slices. 

Since `json.Marshal` yields `null` values for empty slices (see #457), this outcome should be explicit in the schema :slightly_smiling_face: 

## Summary by CodeRabbit

- **New Features**
	- Enhanced schema generation to allow nullable items within arrays.
	- Updated schema definitions to reflect arrays that can contain null values.

- **Bug Fixes**
	- Improved test cases to ensure consistent handling of nullable array types in JSON schema definitions.

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -746,6 +746,7 @@ func schemaFromType(r Registry, t reflect.Type) *Schema {
 			s.ContentEncoding = "base64"
 		} else {
 			s.Type = TypeArray
+			s.Nullable = true
 			s.Items = r.Schema(t.Elem(), true, t.Name()+"Item")
 
 			if t.Kind() == reflect.Array {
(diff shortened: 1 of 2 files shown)


danielgtaylor (approved):
LGTM, thanks!

Also worth mentioning: https://github.com/golang/go/issues/37711 and the json/v2 effort linked there which would address some of this. For now at least the schema will match the behavior.

jzillmann:
Doesn't it depend on the way the slices are initialized ?
Like `emptySlice := []string{}` will marshal into `[]`...
For me it would be nice if there is a configuration switch with which I can steer this behavior. Having a generated typescript client this change is a big pain in the frontend while I haven't been suffering null slices before.

lsdch:
Actually I recently ran into the same issue, also using a client generator. I agree that making this feature optional would be better.
@danielgtaylor I could submit a PR for this, just not sure where this option should belong. What do you think ?

PR #532, opened 2024-08-02
fix(schema): schema example not work in Parameter

fourcels:
```go
type GreetingType int

func (*GreetingType) Schema(r huma.Registry) *huma.Schema {
	schema := &huma.Schema{
		Type:     huma.TypeInteger,
		Examples: []any{1},
	}
	return schema
}

```

## Summary by CodeRabbit

- **New Features**
	- Enhanced example generation logic for improved flexibility and usability in the application.
	- Default examples are now utilized when no explicit examples are defined, enriching the user experience.
	- Introduced a new integer type with an example value, improving API documentation clarity.
- **Tests**
	- Added a test to validate the new integer type's behavior within the API context.


danielgtaylor (commented):
@fourcels this looks great, but can you add a test to cover this use case? Thanks!

danielgtaylor (approved):

PR #542, opened 2024-08-14
fix: race by deferring the return of buf to sync.Pool when using RawBody

nunoo:
You are still using the result of buf.Bytes() in body, later in this function. This defers until completion. Alternatively we can make body a copy of the bytes from the buffer.

https://www.captaincodeman.com/golang-buffer-pool-gotcha#the-gotcha---were-still-using-it

## Summary by CodeRabbit

- **Chores**
	- Improved resource management in the registration process by optimizing buffer handling.
	- Enhanced control flow and error handling through the use of deferred execution for cleanup.

- **Tests**
	- Added a new test function to evaluate race conditions for concurrent API requests, improving the reliability of the API under high load.

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -1292,8 +1292,20 @@ func Register[I, O any](api API, op Operation, handler func(context.Context, *I)
 						}
 					}
 
-					defer bufPool.Put(buf)
-					defer buf.Reset()
+					if rawBodyIndex != -1 {
+						// If the raw body is used, then we must wait until *AFTER* the
+						// handler has run to return the body byte buffer to the pool, as
+						// the handler can read and modify this buffer. The safest way is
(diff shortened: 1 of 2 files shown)


danielgtaylor:
@nunoo thank you for the PR! I added a commit with a test to trigger the issue, and slightly modified your fix to only apply when `RawBody` is in use, so when using a just the normal body struct we get a slight performance improvement by being able to re-use the buffer earlier. This will go into the next release in a few days. :+1:

danielgtaylor (approved):

nunoo:
@danielgtaylor thanks for the quick review and additional context to the changes you made! I'm excited to have contributed to the code base!

Issue #571, opened 2024-09-11
Schema change for arrays has broken API compatibility


harryjph:
Pull request #527, released in version `v2.20.0`, introduced a backwards-incompatible change to the way that Huma generates OpenAPI schemas.

Previously, arrays were recorded as an OpenAPI `array` type, but now they are recorded as either `array` or `null`. This change was pushed out in a minor version update, which the `go` tool will often quietly upgrade to. By upgrading to this version, any application using Huma has subtly made a breaking change to its API schema.

Adding additional types to the schema means that clients generated from the schema now have to add logic to deal with both possibilities. I know that Go can serialise arrays as null, but our application is written in a way such that this is not possible.

This also totally breaks [oapi-codegen](https://github.com/oapi-codegen/oapi-codegen):
```bash
$ oapi-codegen -generate "types,client" -package {package} openapi.json
error generating code: error generating type definitions: error generating Go types for component schemas: error converting Schema {type} to Go type: error generating Go schema for property '{property}': error resolving primitive type: unhandled Schema type: &[string null]
```

There are multiple [other](https://github.com/danielgtaylor/huma/pull/527#issuecomment-2336207081) [people](https://github.com/danielgtaylor/huma/pull/527#issuecomment-2337658017) reporting this issue on the original PR thread.

**Please make the change to how schemas are generated a configurable option.**

harryjph:
Update: I have found that the OpenAPI 3.0 (as opposed to 3.1) schema is not affected by this change, so `oapi-codegen` users should use the `openapi-3.0.json` schema. Closing for now.

PR #594, opened 2024-10-02
Revert nullable schema generation for array slices

lsdch:
1961ef9 reverts #527 which seems to be causing more trouble than good (see #571 and comments on #527).

110fcaa introduces a configuration field in the schema registry, which allows specifying whether schemas for arrays/slices should be nullable by default. This is not ideal since it would rather belong in the [huma.Config](https://pkg.go.dev/github.com/danielgtaylor/huma/v2#Config), but then it would not be accessible at the time of schema generation without major changes in the API. Let me know if you got ideas on how to improve this.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `RegistryConfig` struct to manage nullability of slices in OpenAPI schema generation.
  - Enhanced `Config` methods to provide direct access to registry configurations.

- **Bug Fixes**
  - Improved handling of the `Nullable` property for array types, allowing dynamic control based on configuration settings.

- **Tests**
  - Updated test cases to reflect the new `NullSlices` configuration, ensuring accurate schema generation for nullable arrays and slices.

### DIFF
--- a/api.go
+++ b/api.go
@@ -191,6 +191,10 @@ type Config struct {
 	CreateHooks []func(Config) Config
 }
 
+func (c *Config) RegistryConfig() *RegistryConfig {
+	return c.Components.Schemas.Config()
+}
+
 // API represents a Huma API wrapping a specific router.
 type API interface {
 	// Adapter returns the router adapter for this API, providing a generic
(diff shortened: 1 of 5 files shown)


danielgtaylor:
@lsdch I spent some time thinking about this one and have proposed an alternative in #612. Would you be willing to take a look and let me know what you think?

lsdch:
As commented on #612, I like the more straightforward alternative :+1:  Closing !

PR #612, opened 2024-10-16
fix: make nullable arrays configurable

danielgtaylor:
This is an alternative approach to #594. @lsdch would love to know what you think.

1. There is precedent in Huma to use global variables to control behavior. This adds another such value to set the default array nullability called `huma.DefaultArrayNullable`.
2. The default stays `true`, as this is more correct based on how Go marshals JSON. A `nil` slice will encode as `null` so we ask the user to opt-in to alternative behavior.
3. Updates docs to reflect this and adds a note about `nil` slices encoding as `null`.
4. Adds a few tests.

IMO this prevents another break, moving back to less correct/accurate behavior, and makes everything explicit. I'm not the biggest fan of global variables but it seems reasonable here instead of trying to plumb everything through the various types and interfaces. Thoughts?

## Summary by CodeRabbit

- **Documentation**
	- Updated request validation documentation to clarify nullable fields and rules for optional and required fields.
	- Enhanced examples for making fields nullable in Go structs.

- **New Features**
	- Introduced a global variable to control the default nullability of arrays, simplifying configuration.

- **Tests**
	- Added new test cases for nullable and non-nullable arrays, improving schema validation coverage.

### DIFF
--- a/docs/docs/features/request-validation.md
+++ b/docs/docs/features/request-validation.md
@@ -57,13 +57,14 @@ Huma tries to balance schema simplicity, usability, and broad compatibility with
 Fields being nullable is determined automatically but can be overridden as needed using the logic below:
 
 1. Start with no fields as nullable
-2. If a field is a pointer:
+2. If a field is a pointer (including slices):
     1. To a `boolean`, `integer`, `number`, `string`: it is nullable unless it has `omitempty`.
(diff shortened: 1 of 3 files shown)


lsdch:
Looks good to me ! I like the idea of keeping things simple by using a global variable in this case :+1:

PR #628, opened 2024-10-26
fix: set defaults during validation to prevent override zero values

barata0:
## Release Notes

- **New Features**
  - Enhanced input validation with default value handling across various components.
  - Introduced new input types with resolver logic for improved error handling.

- **Bug Fixes**
  - Improved error handling for input validation failures, ensuring clearer error messages.

- **Tests**
  - Added new test cases to verify the handling of default values in request bodies.
  - Updated existing tests to utilize the new validation method, ensuring comprehensive coverage.

These changes enhance the robustness and reliability of input handling within the application.

Issue #627, opened 2024-10-25
When the tag `default` is used in a bool field huma is ignoring the sent value


robsonpeixoto:
```go
{
	Name: "request-body-defaults-respect-input",
	Register: func(t *testing.T, api huma.API) {
		huma.Register(api, huma.Operation{
			Method: http.MethodPut,
			Path:   "/body",
		}, func(ctx context.Context, input *struct {
			Body struct {
				// Test defaults for primitive types.
				Name  string `json:"name,omitempty" default:"Huma"`
				Count int    `json:"count,omitempty" default:"5"`
				// Test defaults for slices of primitives.
				Tags    []string `json:"tags,omitempty" default:"foo, bar"`
				Numbers []int    `json:"numbers,omitempty" default:"[1, 2, 3]"`
				// Test defaults for fields within slices of structs.
				Items []struct {
					ID       int  `json:"id"`
					Verified bool `json:"verified,omitempty" default:"true"`
				} `json:"items,omitempty"`
			}
		},
		) (*struct{}, error) {
			assert.Equal(t, "Huma", input.Body.Name)
			assert.Equal(t, 5, input.Body.Count)
			assert.Equal(t, []string{"foo", "bar"}, input.Body.Tags)
			assert.Equal(t, []int{1, 2, 3}, input.Body.Numbers)
			assert.Equal(t, 1, input.Body.Items[0].ID)
			assert.False(t, input.Body.Items[0].Verified)
			return nil, nil
		})
	},
	Method: http.MethodPut,
	URL:    "/body",
	Body:   `{"items": [{"id": 1, "verified": false}]}`,
},

```

For the body `{"items": [{"id": 1, "verified": false}]}`, the field `Verified` should be false, but the `default:"true"` is forcing the value to true.

barata0:
[https://github.com/danielgtaylor/huma/pull/597/files](https://github.com/danielgtaylor/huma/pull/597/files) fix this issue?

barata0:
It is not only with bool fields.
But every field that responds true to IsZero() (eg. Numeric Zero count=0) will be overridden by defaults.
[This PR](https://github.com/danielgtaylor/huma/pull/628) fixes this

danielgtaylor:
Fixed in #633.

PR #633, opened 2024-10-31
feat: allow scalar pointers with defaults

danielgtaylor:
This PR is an attempt to fix a long-standing issue where defaults would override explicitly sent zero values from a client if the zero value has semantic meaning for an operation handler.

It implements a new feature that allows the use of scalar pointers with defaults to match the existing behavior of the standard library JSON and other marshaling packages like https://github.com/fxamacker/cbor, which is to use a scalar pointer to determine whether a value has been set by a client or not. You can now do something like this:

```go
type MyInput struct {
	Body struct {
		Enabled *bool `json:"enabled,omitempty" default:"true"`
	}
}
```

The behavior is:

| Client sends | Handler sees |
| -------------|-------------- |
| `true` | `true` |
| `false` | `false` |
| `null` / `undefined` | `true` |

This implementation does not require any additional marshal/unmarshal steps or impact the performance of the running service. It does require some additional reflection & conversions on service startup, but that's a one-time cost and the additional code complexity is not too bad.

Just to note, you still cannot provide defaults for pointers to structs. It would be possible to support this but requires different unmarshaling code in `jsonTagValue` so I'm not going to implement it unless there is a good reason to do so.

FYI @Grumpfy, @robsonpeixoto, @barata0 - what do you think?

Fixes #597, #627, #628

## Summary by CodeRabbit

- **Documentation**
	- Expanded and clarified request validation documentation, including details on struct tags, optional/required fields, nullable fields, and validation tags.
	- Added sections on defaults, read/write-only fields, strict vs. loose validation, and advanced validation rules.

- **Bug Fixes**
	- Improved error handling and parameter processing in the API framework, enhancing clarity and robustness.

- **Tests**
	- Enhanced middleware and request handling tests, including scenarios for cookie handling and request bodies with pointers.
	- Added tests for various error conditions and improved OpenAPI documentation generation.

### DIFF
--- a/huma_test.go
+++ b/huma_test.go
@@ -665,7 +665,7 @@ func TestFeatures(t *testing.T) {
 					}
 				}) (*struct{}, error) {
 					assert.EqualValues(t, "Huma", *input.Body.Name)
-					assert.EqualValues(t, true, *input.Body.Enabled)
+					assert.True(t, *input.Body.Enabled)
 					assert.EqualValues(t, []*string{Ptr("foo"), Ptr("bar")}, input.Body.Tags)
 					assert.EqualValues(t, []*int{Ptr(1), Ptr(2), Ptr(3)}, input.Body.Numbers)
 					assert.Equal(t, 1, input.Body.Items[0].ID)
@@ -697,7 +697,7 @@ func TestFeatures(t *testing.T) {
 				}) (*struct{}, error) {
 					// Ensure we can send the zero value and it doesn't get overridden.
 					assert.EqualValues(t, "", *input.Body.Name)
-					assert.EqualValues(t, false, *input.Body.Enabled)
+					assert.False(t, *input.Body.Enabled)
 					assert.Equal(t, 1, input.Body.Items[0].ID)
 					assert.False(t, *input.Body.Items[0].Verified)
 					return nil, nil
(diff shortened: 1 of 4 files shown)


barata0:
Do you think omitting a field and explciti assing a null value should have different ouputs?  
eg.
```Body:   `{"items": [{"id": 1, verified: null}]}`,```
and
```Body:   `{"items": [{"id": 1}]}`,```
?

File: huma_test.go [hunk 1]
```diff
@@ -642,6 +643,70 @@ func TestFeatures(t *testing.T) {
...
+					assert.EqualValues(t, []*string{Ptr("foo"), Ptr("bar")}, input.Body.Tags)
+					assert.EqualValues(t, []*int{Ptr(1), Ptr(2), Ptr(3)}, input.Body.Numbers)
+					assert.Equal(t, 1, input.Body.Items[0].ID)
+					assert.True(t, *input.Body.Items[0].Verified)
+					return nil, nil
+				})
+			},
+			Method: http.MethodPut,
+			URL:    "/body",
+			Body:   `{"items": [{"id": 1}]}`,
```
barata0, line 678: What about testing defaut values in multiple items here?
```
assert.True(t, *input.Body.Items[0].Verified)
assert.False(t, *input.Body.Items[1].Verified)
assert.True(t, *input.Body.Items[2].Verified)

Body:   `{"items": [{"id": 1}, {"id": 2, "verified": false}, {"id": 3}]}`,
```

danielgtaylor:
> Do you think omitting a field and explcity assigning a null value should have different ouputs?

@barata0 since Go's JSON marshaler and the CBOR marshaler we use don't differentiate between those I don't think it it makes sense at the parsing layer. Sending `null` for something with a default will be just like explicitly saying you want the default. I think this is just a limitation of Go we have to deal with, though I'm open to ideas!

barata0:
I agree.  Cool.
I don´t think I can technically review your PR because I'm new in GO and u r using a lot of reflection (I didn't get there yet)
But it looks ok to me!

PR #659, opened 2024-11-22
Propagate fiberCtx.UserContext() to huma.Context - Context()

excavador:
The problem: fiber uses UserContext() to hold user data, Huma invokes Context(). As the result, I could not propagate data inside context from Fiber. I have JWT/some other fiber middleware (this is right for us, we use fiber without Huma also), but Huma.Context.Context does not see these values

fix: adapters/humafiber - fixed Context() method - use 'fiberCtx.Context()' for everything, except 'Value', 'Value' first tries to find value inside 'fiberCtx.UserContext()' and only if value is missed call original'fiberCtx.Context().Value()'

## Summary by CodeRabbit

- **New Features**
	- Introduced a new context adapter that enhances value retrieval by prioritizing user context in the Fiber framework.
	- Updated context handling in the application to utilize the new adapter for improved compatibility with Go's context management.

### DIFF
--- a/adapters/humafiber/humafiber.go
+++ b/adapters/humafiber/humafiber.go
@@ -15,6 +15,35 @@ import (
 	"github.com/gofiber/fiber/v2"
 )
 
+// avoid race condition inside fasthttp need to cache Context().Done() and UserContext().Value
+type contextAdapter struct {
+	*fiber.Ctx
+	done <-chan struct{}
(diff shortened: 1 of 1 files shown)


danielgtaylor (approved):
Thanks, I wasn't aware of `UserContext` in Fiber!

PR #668, opened 2024-12-05
fix oauth2-jwt.md mermaid rendering issue

ssoroka-tc:
mermaid has some problem where `1.` is interpreted as markdown and doesn't render properly. Not sure if that's a recent problem, but it manifests like so:

Screenshot 2024-12-04 at 11 36 41 PM

changing to `1:` avoids this problem.


danielgtaylor (approved):
Thank you! :+1:

Issue #694, opened 2024-12-31
Allow non-FormFile types in multipart forms


b-kamphorst:
I adopted Huma in my project over the last days and I really like it! In general I thought it was quite easy to adopt Huma with the exception of Enums (https://github.com/danielgtaylor/huma/issues/621#issuecomment-2456588788 helped me out) and multipart form data. I found the documentation on multipart form data lacking, but I could get a functioning version after looking at the [tests](https://github.com/danielgtaylor/huma/blob/v2.27.0/huma_test.go#L906).

My web server receives multipart form data that includes both files and strings. I made a change to Huma v2.27.0 that allows string fields within RawData/MultipartFormFiles. I am curious whether this is a meaningful addition to your work? In extention, it could be useful to allow for fields of arbitrary type within MultipartFormFiles?

Looking forward to your thoughts and keep up the good work!

PR #695, opened 2024-12-31
feat: allow string fields in MultipartFormFiles

b-kamphorst:
Introduces ability to document, allow, and use fields of type string within MultipartFormFiles. Validation is limited to checking existence of the field in the form, and ensuring that at most one field with that name exists. In particular, requirements such as `minLength` are not considered.

Relates to #694.


b-kamphorst:
Superseded by https://github.com/danielgtaylor/huma/pull/706.

PR #705, opened 2025-01-09
refactor: split up huma.Register

b-kamphorst:
The current [`huma.Register`](https://github.com/danielgtaylor/huma/blob/v2.27.0/huma.go#L607) function is 900+ lines of code and has [up to 9 levels of nesting](https://github.com/danielgtaylor/huma/blob/v2.27.0/huma.go#L1378). While working on #694 I found it really hard to understand the logic. In my effort to understand the function, I performed several refactorings and I thought that you may appreciate them as well.

Surely some parts are a matter of taste -- I don't mind changing them.

Looking forward to your thoughts.

### DIFF
--- a/formdata.go
+++ b/formdata.go
@@ -85,71 +85,6 @@ func (v MimeTypeValidator) Validate(fh *multipart.FileHeader, location string) (
 	}
 }
 
-func (m *MultipartFormFiles[T]) readFile(
-	fh *multipart.FileHeader,
-	location string,
-	validator MimeTypeValidator,
-) (FormFile, *ErrorDetail) {
-	f, err := fh.Open()
-	if err != nil {
-		return FormFile{}, &ErrorDetail{Message: "Failed to open file", Location: location}
-	}
-	contentType, validationErr := validator.Validate(fh, location)
-	if validationErr != nil {
-		return FormFile{}, validationErr
-	}
-	return FormFile{
-		File:        f,
-		ContentType: contentType,
-		IsSet:       true,
-		Size:        fh.Size,
-		Filename:    fh.Filename,
-	}, nil
-}
-
-func (m *MultipartFormFiles[T]) readSingleFile(key string, opMediaType *MediaType) (FormFile, *ErrorDetail) {
-	fileHeaders := m.Form.File[key]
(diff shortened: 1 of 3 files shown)


b-kamphorst:
The diff is quite large, so I made an effort to clean up the branch history. I think it may be easiest to review this PR by traversing the commits. Every commit passes the tests (`git rebase main -x 'go test ./...'`), as should be expected in a refactoring. 

Before I cleaned up the history I made a change, passed the tests, and later figured out that it broke the flow of logic for the `TextUnmarshaler`. I added a test to prevent that from happening again. This just reminds us that even refactor efforts that pass all tests should be reviewed with care :)

danielgtaylor:
@b-kamphorst this has been on my ToDo list for ages, so 1. Thank you so much for taking the time to do this and 2. Sorry I didn't get to it earlier!

Looks like there is a small conflict I may be able to resolve, but I'll need to take some time and review. Originally I mashed much of this code together in an effort to keep it performant, so I will run some benchmarks too. Give me a couple days please - I do want to merge this cleanup!

b-kamphorst:
No worries, I appreciate your devotion to this wonderful project and I am glad that this may be of value. I'll rebase on `main` to resolve the minor conflict tomorrow. 

Please take your time. Looking forward to your feedback.

danielgtaylor:
@b-kamphorst thanks for all this work! I've reviewed the code and it looks good. The one issue I found is that benchmarks showed an increase in dynamic memory allocations per request and I was able to `git bisect` it to https://github.com/danielgtaylor/huma/pull/705/commits/78b0cc40028ec54d19d89837124c25bf030eb7ad, specifically the change in splitting query param strings before they might be needed. I've pushed a fix to move that logic to only be applied if the underlying parameter is actually an array of values. The performance is now equal to before the PR, so I am going to merge. Thanks again 🥇

danielgtaylor (approved):

b-kamphorst:
That's nice! If you have any recommended sources on (that kind of) Go benchmarking then I would be interested. I'm a couple of months into the language.

Glad to hear you valued the refactoring, happy to help!

PR #714, opened 2025-01-23
fix: implement findRelativeRessourcePath function to fix PatchResourc…

### LINKED ISSUE #713: Autopatch not working when a prefix is present in the path
When a prefix is present in the path, the autopatch feature does not work as expected.
A 404 error occurs when a `PATCH` request is sent to the server.

## Minimal Reproductible Example

```go
func main() {
    router := chi.NewRouter()
    router.Route("/api", apiMux())
    err = http.ListenAndServe(fmt.Sprintf(":%d", 8080), router)
}

// apiMux returns a function that initializes the API routes
func apiMux() func(chi.Router) {
        return func(router chi.Router) {
            humaConfig := huma.DefaultConfig("API", "dev")
            humaConfig = openapi.WithAuthSchemes(humaConfig)
            humaConfig = openapi.WithOverviewDoc(humaConfig)
            humaConfig = openapi.WithServers(humaConfig, config)
            api := humachi.New(router, humaConfig)
                Method:      "GET",
                Path:        "/ressources/{id}",
            }, getRessourceByID)
                Method:      "PUT",
                Path:        "/ressources/{id}",
            }, updateRessourceByID)
            autopatch.AutoPatch(api)
    }
}
```

## Technical details

The probleme is here: <https://github.com/danielgtaylor/huma/blob/7c5c0d3425264809109bfb059162d369fa862620/autopatch/autopatch.go#L185>

Since `ctx.URL().Path` give the path with the prefix, for example `/api/ressources/1`, the `/api` prefix stay in the path given for the request, and the router does not find the route:  <https://github.com/danielgtaylor/huma/blob/7c5c0d3425264809109bfb059162d369fa862620/autopatch/autopatch.go#L214>

## Possible Solution
I made an MR showing a potential solution here #714

froz42:
This merge request fix issue #713 "Autopatch not working when a prefix is present in the path"

## Technical details

I implemented a new function `findRelativeRessourcePath` in the `autopatch` package that removes the prefix from the path before searching for the resource in the router by comparing the `PUT` path and request path.

It will strip the request path until the prefix of the `PUT` path prefix is found.

It think a better solution could be found if with a proper rework.

I wrote unit tests for the new function.

Fixes #713.

### DIFF
--- a/autopatch/autopatch.go
+++ b/autopatch/autopatch.go
@@ -181,8 +181,10 @@ func PatchResource(api huma.API, path *huma.PathItem) {
 			return
 		}
 
+		resourcePath := findRelativeResourcePath(ctx.URL().Path, put.Path)
+
 		// Perform the get!
-		origReq, err := http.NewRequest(http.MethodGet, ctx.URL().Path, nil)
+		origReq, err := http.NewRequest(http.MethodGet, resourcePath, nil)
 		if err != nil {
 			huma.WriteErr(api, ctx, http.StatusInternalServerError, "Unable to get resource", err)
 			return
@@ -281,7 +283,7 @@ func PatchResource(api huma.API, path *huma.PathItem) {
 		}
 
 		// Write the updated data back to the server!
-		putReq, err := http.NewRequest(http.MethodPut, ctx.URL().Path, bytes.NewReader(patched))
+		putReq, err := http.NewRequest(http.MethodPut, resourcePath, bytes.NewReader(patched))
 		if err != nil {
 			huma.WriteErr(api, ctx, http.StatusInternalServerError, "Unable to put modified resource", err)
 			return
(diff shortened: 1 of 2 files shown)


autopatch/autopatch.go
```diff
@@ -181,8 +181,10 @@ func PatchResource(api huma.API, path *huma.PathItem) {
 			return
 		}
 
+		ressourcePath := findRelativeRessourcePath(ctx.URL().Path, put.Path)
```
danielgtaylor, line 184: Minor typo:

```suggestion
		resourcePath := findRelativeResourcePath(ctx.URL().Path, put.Path)
```

Also has to be fixed further down below.

autopatch/autopatch.go
```diff
@@ -400,3 +402,32 @@ func makeOptionalSchema(s *huma.Schema) *huma.Schema {
...
 }
+
+// This function help to find the relative path of the resource to patch
+// this allow to handle potential prefix in the path
+// for example if the requestPath is /api/v1/user/1 and the put path is /user/{id}
+// the function will return /user/1
+func findRelativeRessourcePath(requestPath string, putPath string) string {
+	putPathParts := strings.Split(putPath, "/")
+	// if the path is not deep enough, we return the original path
+	if len(putPathParts) < 2 {
```
danielgtaylor, line 413: What determines this magic number here? Why `2`? Are there any cases where this could fail?

froz42: The number 2 is here to handle this case:

if putPath is `/{id}` ->` len(putPathParts == 1` that mean that we don't have a ressource name to find,

A case that could be not covered by this could be:

GET {prefix}/singleRessource
PUT {prefix}/singleRessource

it also protect the table access `wantedPrefix := putPathParts[1]`

froz42: In the case that is not supported we just fall back to the default behavior

danielgtaylor (approved):

PR #718, opened 2025-02-03
fix: double validation errors

### LINKED ISSUE #717: Invalid boolean query parameters return two errors instead of one since 2.28.0
When trying to upgrade from 2.27.0 to 2.28.0, I noticed a change in validation of boolean query parameters. Taking the example from the README (replacing Chi with default router for simplicity) and adding a query parameter to the request demonstrates this change:

```go
package main

import (
	"context"
	"fmt"
	"net/http"

	"github.com/danielgtaylor/huma/v2/adapters/humago"
	"github.com/danielgtaylor/huma/v2/humacli"
)

type Options struct {
	Port int `help:"Port to listen on" short:"p" default:"8888"`
}

type GreetingOutput struct {
	Body struct {
	}
}

func main() {
	cli := humacli.New(func(hooks humacli.Hooks, options *Options) {
		api := humago.New(http.DefaultServeMux, huma.DefaultConfig("My API", "1.0.0"))

		huma.Get(api, "/greeting/{name}", func(ctx context.Context, input *struct{
			Boolean bool `query:"boolean"` // This is the query parameter added.
			Name string `path:"name" maxLength:"30" example:"world" doc:"Name to greet"`
		}) (*GreetingOutput, error) {
			resp := &GreetingOutput{}
			resp.Body.Message = fmt.Sprintf("Hello, %s!", input.Name)
			return resp, nil
		})

		hooks.OnStart(func() {
			http.ListenAndServe(fmt.Sprintf(":%d", options.Port),nil)
		})
	})

	cli.Run()
}
```

With Huma v2.27.0, running `curl 'localhost:8888/greeting/world?boolean=invalid' | jq` outputs:

```json
{
  "$schema": "http://localhost:8888/schemas/ErrorModel.json",
  "status": 422,
  "errors": [
    {
      "message": "invalid boolean",
      "location": "query.boolean",
      "value": "invalid"
    }
  ]
}
```

However, Huma v2.28.0 replies to the same command with:

```json
{
  "$schema": "http://localhost:8888/schemas/ErrorModel.json",
  "status": 422,
  "errors": [
    {
      "message": "invalid boolean",
      "location": "query.boolean",
      "value": "invalid"
    },
    {
      "message": "expected boolean",
      "location": "query.boolean"
    }
  ]
}
```

It doesn't seem to be an intentional change since it's not documented and it doesn't make much sense to returns the same error twice.

skwair:
### Description

This PR fixes a regression introduced in #705, described in #717.

While hardening test cases, I noticed that slices of uint8s were not supported, so I took the liberty to also add the implementation (in a separate commit though, which can be reverted if this was intentional).

Fixes #717

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -696,6 +696,7 @@ func Register[I, O any](api API, op Operation, handler func(context.Context, *I)
 			pv, err := parseInto(ctx, f, value, *p)
 			if err != nil {
 				res.Add(pb, value, err.Error())
+				return
 			}
 
 			if !op.SkipValidateParams {
--- a/huma.go
+++ b/huma.go
@@ -1414,6 +1414,20 @@ func parseSliceInto(f reflect.Value, values []string) (any, error) {
 		f.Set(reflect.ValueOf(vs))
 		return vs, nil
 
+	case reflect.Uint8:
+		vs, err := parseArrElement(values, func(s string) (uint8, error) {
+			val, err := strconv.ParseUint(s, 10, 8)
+			if err != nil {
+				return 0, err
+			}
+			return uint8(val), nil
+		})
+		if err != nil {
+			return nil, errors.New("invalid integer")
+		}
+		f.Set(reflect.ValueOf(vs))
+		return vs, nil
+
 	case reflect.Uint16:
 		vs, err := parseArrElement(values, func(s string) (uint16, error) {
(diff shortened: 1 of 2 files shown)


danielgtaylor (approved):
Thanks!

PR #722, opened 2025-02-10
feat: custom query parameter types

lsdch:
Hi! 

I need a way to handle null values in query parameters (and apparently I am not the only one #438 :slightly_smiling_face: ) so here is a PR that gives the possibility to wrap query parameter fields with custom types. 

It relies on implementing the `ParamWrapper` interface, which exposes as a `reflect.Value` the internal field that we want to use when parsing request parameters. 
`ParamReactor` (bad name, I know :see_no_evil: ) is another interface that can be implemented to specify a callback when a request parameter was parsed into our custom type.

```go
type ParamWrapper interface {
	Receiver() reflect.Value
}

type ParamReactor interface {
	OnParamSet(isSet bool, parsed any)
}
```

Example usage: 
```go 
type OptionalParam[T any] struct {
	Value T
	IsSet bool
}

// Define schema to use wrapped type
func (o OptionalParam[T]) Schema(r huma.Registry) *huma.Schema {
	return huma.SchemaFromType(r, reflect.TypeOf(o.Value))
}

// Expose wrapped value to receive parsed value from Huma
// MUST have pointer receiver
func (o *OptionalParam[T]) Receiver() reflect.Value {
	return reflect.ValueOf(o).Elem().Field(0)
}

// React to request param being parsed to update internal state
// MUST have pointer receiver
func (o *OptionalParam[T]) OnParamSet(isSet bool, parsed any) {
	o.IsSet = isSet
}
```

I believe it resolves validation issues we had with #438, but there might be other caveats I am not aware of. Also I am not 100% convinced with the proposed API, what do you think ?

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -102,6 +102,10 @@ func findParams(registry Registry, op *Operation, t reflect.Type) *findResult[*p
 			Type: f.Type,
 		}
 
+		if reflect.PointerTo(f.Type).Implements(reflect.TypeFor[ParamWrapper]()) {
+			pfi.Type = reflect.New(f.Type).Interface().(ParamWrapper).Receiver().Type()
+		}
+
 		if def := f.Tag.Get("default"); def != "" {
 			pfi.Default = def
 		}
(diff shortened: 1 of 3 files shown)


huma.go
```diff
@@ -696,12 +696,24 @@ func Register[I, O any](api API, op Operation, handler func(context.Context, *I)
 				return
 			}
 
-			pv, err := parseInto(ctx, f, value, nil, *p)
+			var receiver = f
+			if f.Addr().Type().Implements(reflect.TypeFor[ParamWrapper]()) {
+				receiver = f.Addr().Interface().(ParamWrapper).Receiver()
+				// Feels like a hack, but we need to override the type of the wrapper from the cache
+				// There might a be a better way
+				p.Type = receiver.Type()
```
danielgtaylor, line 704: I suspect we want to find the right type in `findParams` at https://github.com/danielgtaylor/huma/blob/main/huma.go#L102 by calling `Receiver()` once on a dummy value, then we don't have to do this check/swap on each request.

lsdch:
Thanks for the feedback @danielgtaylor ! And for the pointer to fix the type precomputation, feels way better now :slightly_smiling_face:

danielgtaylor (approved):

PR #725, opened 2025-02-12
fixed work with fiber.Ctx, fiber.UserContext - fixed graceful shutdown and race condition on access to huma.Context outside handler

excavador:
Fiber and FastHttp has the very strict rules how to work with Context() and UserContext()

If some huma-based application, whatever reason, has references to context.Context outside of Huma handler, then fiber/fasthttp RequestCtx does not work as expected, the most likely you receive crash or race-condition

To prevent race conditions and possible issues with graceful shutdown of the servers, this pull request completely fix the way on how fiber and huma integrated.

It also include golang tests to detect race conditions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,7 +20,7 @@ jobs:
         uses: golangci/golangci-lint-action@v6
         with:
           version: v1.60.1
-      - run: go test -coverprofile=coverage.txt -covermode=atomic ./...
+      - run: go test -race -coverprofile=coverage.txt -covermode=atomic ./...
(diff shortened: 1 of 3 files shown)


excavador:
@danielgtaylor I hope this it the final fix of "humafiver" provider.
I am sorry for several iterations, but "zero-allocation" behavior of fasthttp / fiber is not something so simple to prepare properly

In this Pull Request I have the reproduction case to illustrate issues with fasthttp/fiber in case when user-side code ignores recommentations about fiber/fasthttp (and without penalties for that!)

danielgtaylor (approved):

Issue #670, opened 2024-12-05
Is there a way to register custom Schemas without registering an operation?


Blinkuu:
I'm implementing a WebSockets API and would like to include my custom message types as part of my OpenAPI specification so it's easier for clients to consume.

How could I do it with Huma?

lazharichir:
We use a custom namer for some reasons (we added a struct tag to hardcode, sometimes, the name of the schema) so here is how I do it:

```golang
package openapiregistry

var Registry = huma.NewMapRegistry("#/components/schemas/", MyCustomSchemaNamer)

func init() {
	Registry.Map()["AnyAction"] = &huma.Schema{Type: "string", Enum: actions.AllAsAny()}
	// ... more of these ...
}

func MyCustomSchemaNamer(t reflect.Type, hint string) string {
	// our custom namer, there's a default one you can use
}
```


``` golang
// when instantiating our API
humaConfig := huma.DefaultConfig("Name of our API", "1.0.0")
humaConfig.OpenAPI.Components.Schemas = openapiregistry.Registry
```

I understand you could use the default registry with something like:

```golang
humaConfig.OpenAPI.Components.Schemas.Map()["AnyAction"] = &huma.Schema{Type: "string", Enum: actions.AllAsAny()} 
```

danielgtaylor:
@Blinkuu, thanks for the question. @lazharichir is right you can manually add schemas like in his example. Here is one you can run showing how it works: https://go.dev/play/p/DRcb_umQhX1. Hope this helps!

Blinkuu:
Thanks, I got it working, closing the issue :)

Issue #649, opened 2024-11-12
Implement better deduping of schema names when using anonymous structs for request bodies


mzhu22:
Hi! 

I ran into a `panic: duplicate name: Request ...` error when generating the OpenAPI spec in my API. I noticed this happens when I have >1 POST endpoint taking a request body, where the handler defines the input type via an anonymous struct.

e.g..
```
func handler1(ctx context.Context, input *struct{ Body {...} }) { ... }
func handler2(ctx context.Context, input *struct{ Body {...} }) { ... }
```

It seems like the issue is that, when using anonymous structs for request bodies, the OpenAPI schema for the request body gets a generic name "Request". When there's more than one, a collision occurs.

Would it be possible to make a unique dynamic name, e.g., by prefixing with the endpoint path, name, etc?

I know that using named structs is an easy workaround, but I think anonymous structs work well for these input types, since they are not passed around or initialized in other code. It keeps the code concise and avoids another thing to name and rename!

Thanks! Been really enjoying this package. You're doing great work.

danielgtaylor:
@mzhu22 thanks for opening an issue. Do you have an example of how you are actually running into this problem? Here's an example with anonymous structs that doesn't panic: https://go.dev/play/p/LC0bGxGe1NI

This is a surprisingly complex problem you've wandered into! The current behavior is a deliberate design choice to keep the implementation both simple and reproducible without breaking generated clients (something we ran into with Huma v1 which used unique generated suffixes). Part of the problem is that the order of processing structs can be dependent on a number of things, like the Go version or new structs for new operations that would shift the two that were previously there. For a concrete example, imagine:

```go
type First struct {
  Body struct { /* ... */ }
}

type Second struct {
  Body struct { /* ... */ }
}
```

You would presumably generate `RequestBody1` for `First` and then `RequestBody2` for `Second`. Your generated SDK would have a call like `client.FirstOperation(ctx, sdk.RequestBody1{ /* ... */ })`. Now you introduce:

```go
type Another struct {
  Body struct { /* ... */ }
}
```

Since that may get processed before `First`, you now get `RequestBody1` meaning `Another.Body` instead of `First.Body`, breaking the generated SDK. Your users would have to update to using `client.FirstOperation(ctx, sdk.RequestBody2{ /* ... */ })`.

[...]

PR #731, opened 2025-02-18
feat: add example for type renaming

shakhzodkudratov:
I faced #649 #670 like issues when I have internal `User` struct and conflicting recursive `github.User` through `github.Repository`. Added an example for workaround.


danielgtaylor (approved):
Great example thank you!

PR #736, opened 2025-02-19
feat: support context unwrapping

danielgtaylor:
This PR is an alternative to #699 which enables unwrapping a `huma.Context` back into its router-specific request/response object or router-specific context, effectively escaping Huma and giving back full control to the user. It's generally recommended to not do this, but there are legitimate use-cases where this could be handy to have as an option. Example:

```go
router := http.NewServeMux()
api := humago.New(router, huma.DefaultConfig("My API", "1.0.0"))

api.UseMiddleware(func(ctx huma.Context, next func(huma.Context)) {
	r, w := humago.Unwrap(ctx)
	fmt.Println("Request started", r.Method, r.URL)
	w.Header().Set("X-Request-Start", time.Now().Format(time.RFC3339))
	next(ctx)
})
```

This implementation does not modify any interfaces in Huma itself, nor does it require manually casting `any` or things like that. Each adapter simply provides a convenience `Unwrap()` function that returns the underlying data.

This should also make it fairly simple to switch routers as you can find/replace the router-specific package (e.g. `humachi` → `humago`) and things will just work or provide a compile-time error if the return types changed, making it easy to fix, instead of waiting for a runtime error.

### DIFF
--- a/adapters/humabunrouter/humabunrouter.go
+++ b/adapters/humabunrouter/humabunrouter.go
@@ -20,6 +20,15 @@ import (
 // form data.
 var MultipartMaxMemory int64 = 8 * 1024
 
+// Unwrap extracts the underlying HTTP request and response writer from a Huma
+// context. If passed a context from a different adapter it will panic.
+func Unwrap(ctx huma.Context) (bunrouter.Request, http.ResponseWriter) {
+	if c, ok := ctx.(*bunContext); ok {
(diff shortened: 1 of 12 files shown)


excavador:
@danielgtaylor I am fine with these changes, humefiber adapter will work well (as long as nobody will capture *fiber.Ctx or underlying structures outside of the handlers)

PR #742, opened 2025-02-26
Add 'omitzero' support to mark optional fields

JasirZaeem:
Add support for using the new `omitzero` field tag introduced in go-1.24 to mark those fields as optional/not-required

https://tip.golang.org/doc/go1.24#encodingjsonpkgencodingjson
> When marshaling, a struct field with the new omitzero option in the struct field tag will be omitted if its value is zero.

### Changes
- Added a new check for `omitzero` to mark field as optional in `schema.go`
- Added new tests in to check `omitempty` and `omitzero` in `schema_test.go`
- Mentioned the new tag in `request-validation.md` docs

Lmk if there are other files/places that need changing for this change.

### DIFF
--- a/docs/docs/features/request-validation.md
+++ b/docs/docs/features/request-validation.md
@@ -40,10 +40,10 @@ type MyStruct struct {
 
     // The following are all optional.
     Optional1 string  `json:"optional1,omitempty"`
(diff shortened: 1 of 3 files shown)


danielgtaylor (approved):
Awesome thank you @JasirZaeem :+1:

Issue #560, opened 2024-08-26
Primitive-based types are not reused in the generated OAS.


superstas:
Hey there!

I've got the following types

```go

var _ huma.SchemaTransformer = CustomHeader("")

type CustomHeader string

func (h CustomHeader) TransformSchema(r huma.Registry, s *huma.Schema) *huma.Schema {
	s.Type = "string"
	s.Description = "Custom header Description. The CustomHeader schema should be reused across all endpoints that use it."
	return s
}

type CreateUser struct {
	Body struct {
		Name string `json:"name" doc:"The name."`
	}

	CustomHeader CustomHeader `header:"X-Custom-Header"`
}

type UpdateUser struct {
	Body struct {
		Name string `json:"name" doc:"The name."`
	}

	CustomHeader CustomHeader `header:"X-Custom-Header"`
}

type Output struct {
	Body struct {
		Message string `json:"message" doc:"The message."`
	}
}
```

With the following routes
```go
	huma.Post(api, "/user", func(ctx context.Context, req *CreateUser) (*Output, error) {
		resp := &Output{}
		resp.Body.Message = "It works!"
		return resp, nil
	})

	huma.Patch(api, "/user/{id}", func(ctx context.Context, req *UpdateUser) (*Output, error) {
		resp := &Output{}
		resp.Body.Message = "It works!"
		return resp, nil
	}}
```

I expect that `CustomHeader` would be re-used via `$ref` in the generated OAS.

Instead the `CustomHeader` schema is embedded into all the endpoints

```yaml
paths:
  /user:
    post:
      operationId: post-user
      parameters:
        - description: Custom header Description. The CustomHeader schema should be reused across all endpoints that use it.
          in: header
          name: X-Custom-Header
          schema:
            description: Custom header Description. The CustomHeader schema should be reused across all endpoints that use it.
            type: string
      requestBody:
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/CreateUserBody"
        required: true
      responses:
        "200":
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/OutputBody"
          description: OK
        default:
          content:
            application/problem+json:
              schema:
                $ref: "#/components/schemas/ErrorModel"
          description: Error
      summary: Post user
  /user/{id}:
    patch:
      operationId: patch-user-by-id
      parameters:
        - description: Custom header Description. The CustomHeader schema should be reused across all endpoints that use it.
          in: header
          name: X-Custom-Header
          schema:
[...]

PR #758, opened 2025-03-12
#560: Add reusing of primitive-based types in the generated OAS.

superstas:
Hey @danielgtaylor 

This PR proposes a solution for the issue #560.

We tested this change internally, and it decreased the size of our generated OAS by ~22%.

Conversely, this change may look backward-incompatible ( inlined types are being replaced by `$ref` ), so I implemented it as the registry option.

Could you please look and share your thoughts on what could be improved or what I overlooked?

Thank you.

### DIFF
--- a/registry.go
+++ b/registry.go
@@ -59,13 +59,23 @@ func DefaultSchemaNamer(t reflect.Type, hint string) string {
 	return name
 }
 
+type mapRegistryOption func(*mapRegistry)
+
+// WithPrimitiveTypeReuse allows primitive types to be reused.
+func WithPrimitiveTypeReuse() mapRegistryOption {
+	return func(r *mapRegistry) {
+		r.reusePrimitive = true
+	}
+}
+
 type mapRegistry struct {
-	prefix  string
-	schemas map[string]*Schema
-	types   map[string]reflect.Type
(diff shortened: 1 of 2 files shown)


superstas:
@danielgtaylor, an approach with a field tag looks good; I'll try implementing it in a separate commit.

As for your question:

> How would this get represented in the JSON schema if you are creating refs?

Well, according to the spec ([ref](https://spec.openapis.org/oas/v3.1.0.html#fixed-fields-18)), there can only be the `summary` and `description` fields. A situation like this is a conflict that should be shown to the user in some way. 

Possible solutions:
- Use `$ref` only for the very 1st field and keep the rest untouched
- Keep all the conflict cases untouched

WDYT?

danielgtaylor:
@superstas I'm not sure how to solve this honestly. Maybe if there is additional validation on the struct field via tags then you just cannot use a `$ref` and must generate it inline. So maybe a `$ref` is only generated when:

1. You use a named type
2. There is no additional validation (`doc` should be okay, maybe `example` is okay?)

So if you want to use validation your custom primitive type needs to implement `huma.SchemaProvider` or `huma.SchemaTransformer` and your struct fields shouldn't add any validation.

PR #760, opened 2025-03-13
feat: adding the head method to the list of convenience wrappers

hiddenmarten:
Hi,

Thank you again for your work on this project!

I just noticed that the `head` method is missing here. Hopefully, this small addition will be helpful to others. :)


danielgtaylor (approved):
Thanks! :+1:

PR #773, opened 2025-03-21
fix: Allow explicitely setting the "example" tag to the empty string

mumu1000:
This simple patch allows to explicitly set the "example" tag of fields in both params an schemas to the empty string "" .


danielgtaylor (approved):
Thanks! :+1:

PR #775, opened 2025-03-21
proposal: support docs on separate adapter

costela:
This allows exposing the spec and docs on a separate adapter, potentially under a separate path and/or middleware stacks.

E.g.: (assuming `internalRouter` and `mainRouter` are two `chi.Router`)
```go
opts := huma.DefaultConfig("foo", "0.0.1")
opts.DocsAdapter = humachi.NewAdapter(internalRouter)
opts.CreateHooks = []func(huma.Config) huma.Config{
	huma.DefaultSchemaLinkHook(huma.DefaultSchemaRefPrefix, internalPrefix), // assuming internalRouter above is mounted under internalPrefix
}

adapter := humachi.NewAdapter(mainRouter)

huma.NewAPI(opts, adapter)
```

This is a bit of a WIP:
- I'm not 100% sure the removal of `getAPIPrefix` call in the stoplight template is really as harmless as I think
- the other remaining `getAPIPrefix` call will currently break the schema links if the API  is mounted somewhere other than the root.

But regardless of the limitations above: WDYT about the general idea?

### DIFF
--- a/adapters/humachi/humachi_test.go
+++ b/adapters/humachi/humachi_test.go
@@ -322,6 +322,9 @@ func TestChiRouterPrefix(t *testing.T) {
 	mux.Route("/api", func(r chi.Router) {
 		config := huma.DefaultConfig("My API", "1.0.0")
 		config.Servers = []*huma.Server{{URL: "http://localhost:8888/api"}}
+		// config.CreateHooks = []func(huma.Config) huma.Config{
+		// 	huma.DefaultSchemaLinkHook("/api"),
+		// }
 		api = New(r, config)
 	})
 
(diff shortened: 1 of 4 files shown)


costela:
Actually, now that I look at the code, it looks like `NewSchemaLinkTransformer`'s `prefix` parameter isn't even being used? :see_no_evil:

I've pushed a second take at this. The new one uses the first parameter to `NewSchemaLinkTransformer`, so it _may_ be considered a breaking change?  
Also I've tried to keep respecting the prefix detection via the `OpenAPI.Servers`, though I cannot find any place in the documentation where that's actually recommended? Maybe we can skip this backward compatibility and document explicitly how to mount stuff below the root path? There's a commented block in `TestChiRouterPrefix` that shows the two possible variants.

wolveix:
Hey @costela! Thanks for this contribution, and sorry that it hasn't been reviewed yet! I'm still working on getting up-to-speed with some of the more complex parts of the codebase, but want to try and resolve issues and merge PRs where feasible to do so.

This PR may require @danielgtaylor's review, but in the meantime I'd appreciate it if you could resolve any merge conflicts :)

PR #777, opened 2025-03-24
feat(errors): ErrorModel implement errors.Unwrap

burgesQ:
Allow compatibility for errors returned by huma.NewError* with the std' errors.As/errors.Is


error.go
```diff
@@ -107,8 +112,11 @@ func (e *ErrorModel) Error() string {
 //		Value: 5
 //	})
 func (e *ErrorModel) Add(err error) {
+	e.errors = append(e.errors, err)
```
danielgtaylor, line 115: This is great, but `huma.NewError` doesn't actually call `Add` on the error when constructing the details, and the same is potentially the case when manually returning error models in other code (and handlers in your service). I wonder if there is a better way to support most uses instead of requiring `Add` to get called 🤔

burgesQ: Maybe smth like errors.Joins on already existing  `.Errors` field ?

burgesQ: Hehe I think we're good !
My changes were for some code that did some stuff like  `return nil, huma.Error500...("error", ErrorSpecialType{err})`. Turn out my "middleware" tests pass if I change it to smth like `return nil, ErrorSpecialType{huma.Error500...("error", err)`

PR #788, opened 2025-04-04
Add WithIndex to PathBuffer

chrisguitarguy:
Same as `With()`, but takes and int, calls `PushIndex`, converts to a string, then pops.

Found myself wishing for this functionality today, but not a deal breaker if it doesn't make it in. Just a little more convenient.

I did not see any tests for pathbuffer itself, but since this method wouldn't be covered elsewhere I added a test for it.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -129,6 +129,18 @@ func (b *PathBuffer) With(s string) string {
 	return tmp
 }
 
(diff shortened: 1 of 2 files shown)


danielgtaylor:
Thanks!

Issue #634, opened 2024-11-01
Autopatch does not work with nullable fields


betaprior:
Consider a simple struct with two nullable string fields:
```
type Foo struct {
	A *string `json:"a"`
	B *string `json:"b"`
}
```
It is evidently impossible to set these fields to `nil` via a `PATCH` with `application/merge-patch+json` content type. For instance, the following request will throw a 422 with a message "expected required property a to be present".
```
curl --request PATCH \
  --url http://localhost:8888/foo/1 \
  --header 'Accept: application/json, application/problem+json' \
  --header 'Content-Type: application/merge-patch+json' \
  --data '{
  "a": null
}'
```

This seems like a bug in merge patch computation: the `PUT` body in the autogenerated `GET`/`PUT` sequence omits the `a` field altogether instead of setting `a: null` as expected.

The only way to set `a` to `nil` appears to be via a `PUT` of the full struct or via a `json-patch` (not `merge-patch`):
```
curl --request PATCH \
  --url http://localhost:8888/foo/1 \
  --header 'Accept: application/json, application/problem+json' \
  --header 'Content-Type: application/json-patch+json' \
  --data '[
  { "op": "replace", "path": "/a", "value": null }
]
```
This came as an unfortunate surprise as our front-end client started making use of PATCH requests.

danielgtaylor:
@betaprior I dug into this a bit. I originally thought this might be a bug but now I'm not so sure. I re-read RFC 7386 JSON Merge Patch, specifically:

```
define MergePatch(Target, Patch):
  if Patch is an Object:
    if Target is not an Object:
  Target = {} # Ignore the contents and set it to an empty Object
    for each Name/Value pair in Patch:
  if Value is null:
    if Name exists in Target:
      remove the Name/Value pair from Target
  else:
    Target[Name] = MergePatch(Target[Name], Value)
    return Target
  else:
    return Patch
```
https://datatracker.ietf.org/doc/html/rfc7386#section-2

This matches my understanding of Huma's autopatch implementation, and adding in some debug statements I'm able to see that the resulting request made to the `PUT` endpoint indeed has the `a` field removed, which in turn causes the validation error because the field is required.

I'm curious what you think the correct behavior would be in this case.

One thing that you can do is make the field optional. For example this works fine: https://go.dev/play/p/EJmLwkLJ4ci

See also https://huma.rocks/features/request-validation/#optional-required

betaprior:
[...]

PR #791, opened 2025-04-07
Make autopatch merge-patch work with nullable fields

betaprior:


danielgtaylor:
@betaprior can you maybe describe why this is needed? I've looked at it a couple times now and am struggling to fully understand this PR. It doesn't comply with https://datatracker.ietf.org/doc/html/rfc7386 and seems to set a sentinel value, but I wonder if this would better be handled using a middleware of some sorts or by using a different patch mechanism than JSON Merge-Patch. For example:

- [JSON Patch](https://jsonpatch.me/) `{"op": "replace", "path": "/a", "value": null}, {"op": "remove", "path": "/b"}`
- [Shorthand patch format](https://github.com/danielgtaylor/shorthand?tab=readme-ov-file#patch-partial-update) `a: null, b: undefined`)

I think the JSON Merge-Patch implementation should probably stick to the RFC standard. What do you think?

betaprior:
@danielgtaylor we discussed this in https://github.com/danielgtaylor/huma/issues/634. In that thread, you said 

> I would consider copying the autopatch code and modifying it slightly to leave the null values in place for the PUT operation. You don't have to exactly follow JSON merge patch - just make it useful for you and your clients.

That's exactly what this PR implements. Notably, it's not an approach that I originally wanted to take for the exact reason that you mentioned -- I don't love the idea of going against RFC standards, and we experimented with using the shorthand patch format, but at the end decided that sending JSON bodies with `undefined` is even worse with respect to established standards.

Note that the behavior in this PR isn't changing JSON Merge-Patch implementation -- it's provided by the 3rd party library -- but rather mutates the input and the output to the merge patch function so that it results in desired behavior with respect to nulls (i.e., preserving them instead of dropping them). 

If you think this can be done in middleware, could you outline how? The current approach needs the input and output of the patch function, and offhand I don't think it's possible to access in middleware -- unless I suppose the middleware itself somehow intercepts patch requests and provides its own patch implementation.

danielgtaylor:
Going through this again, I see what you are saying about using a middleware. Any other way would be less efficient than doing this right in the patch handler. I think this is alright to merge as an optional extension and it's a well understood concept to use sentinel values with a special meaning. I don't *love* it but I can see the value, it provides an alternative people can use, and the implementation is clean. Thanks!

PR #792, opened 2025-04-08
proposal: extend netip.Addr support to ipv6

costela:
[netip.Addr](https://pkg.go.dev/net/netip#Addr) supports parsing both ipv4 and ipv6 out of the box. So it is a bit surprising to have a `netip.Addr` field in an API only support ipv4 and no easy way to add "dual-stack" support for APIs.

I propose to extend the support for `netip.Addr` added in !396 and support ipv6 out of the box as well.

⚠️ The change in the `format` detection for `netip.Addr` fields may be considered breaking, since existing APIs will start accepting IPv6 addresses and people may be counting on that limitation. I'd argue that it's worth it, but if not, we can remove that single change, while still adding support for the `ip` dual-stack format, to be used explicitly.

WDYT?

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -6,6 +6,7 @@ import (
 	"math"
 	"net"
 	"net/mail"
+	"net/netip"
 	"net/url"
 	"reflect"
 	"regexp"
@@ -234,6 +235,10 @@ func validateFormat(path *PathBuffer, str string, s *Schema, res *ValidateResult
 	case "ipv6":
 		if ip := net.ParseIP(str); ip == nil || ip.To16() == nil {
 			res.Add(path, str, validation.MsgExpectedRFC2373IPv6)
+		}
+	case "ip":
+		if _, err := netip.ParseAddr(str); err != nil {
+			res.Add(path, str, validation.MsgExpectedRFCIPAddr)
 		}
 	case "uri", "uri-reference", "iri", "iri-reference":
 		if _, err := url.Parse(str); err != nil {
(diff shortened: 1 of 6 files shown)


costela:
Taking a second look at this, it appears it _is_ possible to support dual-stack input by setting `format:"ipv6"`, because [IP.To16](https://pkg.go.dev/net#IP.To16) also works for ipv4.

That doesn't seem very intuitive and could maybe warrant a doc fix?

Personally I'd try making this a bit more explicit: the `ipv4`/`ipv6` formats should _really_ be specific to their respective IP versions, and a new `ip` format should support both. WDYT?

Also, it may be a good idea to switch to `netip` for all validations regardless, just for the micro-optimization:

```
pkg: github.com/danielgtaylor/huma/v2
cpu: 12th Gen Intel(R) Core(TM) i7-1250U
BenchmarkNetParseIP-12               31314682                32.42 ns/op           16 B/op          1 allocs/op
BenchmarkNetipParseAddr-12           72398744                15.65 ns/op            0 B/op          0 allocs/op
```

validate.go
```diff
@@ -235,6 +236,10 @@ func validateFormat(path *PathBuffer, str string, s *Schema, res *ValidateResult
 		if ip := net.ParseIP(str); ip == nil || ip.To16() == nil {
 			res.Add(path, str, validation.MsgExpectedRFC2373IPv6)
 		}
+	case "ip":
+		if _, err := netip.ParseAddr(str); err != nil {
+			res.Add(path, str, validation.MsgExpectedRFCIPAddr)
+		}
```
danielgtaylor, line 242: Love this! One concern I have is that `ipv4` and `ipv6` are actually in the JSON Schema spec but `ip` is not. See https://json-schema.org/understanding-json-schema/reference/type#ip-addresses.

I wonder if we should support all three? The two standard ones, and then `ip` for either one since Go supports it well. What do you think?

thushjandan:
Would be awesome if this PR could be merged

wolveix:
Awesome work, thanks so much @costela! Sorry this sat for so long :)

PR #799, opened 2025-04-13
minor: use ErrUnknownContentType in more places

costela:
This small PR just extends the usage of ErrUnknownContentType to a couple more places where the error was not being wrapped and the error string was being duplicated.

AFAICT that shouldn't have any negative side-effects, except for making error testing a bit more consistent?


danielgtaylor (approved):
Thanks! :+1:

PR #806, opened 2025-04-18
fix: make order of SSE events stable in openapi

### LINKED ISSUE #797: SSE definition introduces unstable order in openapi.yaml
I have this SSE setup:
```
sse.Register(api, huma.Operation{
		OperationID:   "StatusEvents",
		Method:        http.MethodGet,
		Path:          path,
		Summary:       "Get status updates",
		DefaultStatus: 200,
	}, map[string]any{
		// Mapping of event type name to Go struct for that event.
		"status":       util.StatusMessage{},
		"progress":     util.ProgressUpdate{},
		"progressData": util.ProgressData{},
		"errorMsg":     util.ErrorMessage{},
	}, func(ctx context.Context, input *schemas.UUIDInput, send sse.Sender) {

```

the `openapi.yaml` which I generate and put under version control is now having meaningless ordering changes every now and then:

Image

culli:
fixes #797

### DIFF
--- a/sse/sse.go
+++ b/sse/sse.go
@@ -9,6 +9,8 @@ import (
 	"os"
 	"reflect"
 	"runtime/debug"
+	"slices"
+	"strings"
 	"time"
 
 	"github.com/danielgtaylor/huma/v2"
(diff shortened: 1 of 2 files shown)


danielgtaylor:
Thanks! :+1:

PR #807, opened 2025-04-22
Allow empty path in group operations

Serjlee:
Hello! I recently found out the validation of the `path` parameter prevented the use of an empty string in the path.

In my use case I found myself in this situation:
- I wanted to create a `/resource` API Group, with CRUD for `resource`, plus some sets of endpoints for sub-resources (ie: `/resource/:id/stuff`)
- I got a `StripSlashes` middleware (required to handle legacy clients), which trims any `/` from incoming requests
- Since I have `StripSlashes`, If I want to add a  `POST /resource/` I need to actually define the route as `POST /resource`
- The above is not possible because of the strict validation done in `Register`
   https://github.com/danielgtaylor/huma/blob/6bfeebae55ed05e1e909041c5e3692dc168cd066/huma.go#L625-L626

With this PR I relaxed the validation of the `Path` parameter, allowing the empty string in the case of operations defined for a Group with at least a prefix defined

### DIFF
--- a/group_test.go
+++ b/group_test.go
@@ -26,6 +26,24 @@ func TestGroupNoPrefix(t *testing.T) {
 	assert.Equal(t, http.StatusNoContent, resp.Result().StatusCode)
 }
 
+func TestGroupEmptyPath(t *testing.T) {
+	_, api := humatest.New(t)
+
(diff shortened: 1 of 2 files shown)


danielgtaylor (approved):
Thanks :+1:

PR #810, opened 2025-04-29
Add option to customize the Gorilla Mux route

RangelReale:
This MR adds an option to the Gorilla mux adapter, allowing customizing the mux route.

This allows, for example, adding standard HTTP handlers per route, maybe using `huma.Operation.Metadata` as input parameters.

### DIFF
--- a/adapters/humamux/humagmux_test.go
+++ b/adapters/humamux/humagmux_test.go
@@ -2,6 +2,7 @@ package humamux
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


RangelReale:
my usecase is using `http.TimeoutHandler` per route, it changes the ResponseWriter, so it is impossible to do with Huma middleware.

danielgtaylor (approved):
Thanks! :+1:

PR #813, opened 2025-05-03
feat: Tag instructed recursive params finding

### LINKED ISSUE #811: How to create a wrapper func that also requires the *http.Request object?
Hi there, I'm trying to create a resuable wrapper function that can do some common logic against the `*http.Request`. Below is my attempt:
```go
// RequestAware will try to capture the http.Request from the huma.Context by implementing the huma.Resolver interface
type RequestAware struct {
	*http.Request
}

func (i *RequestAware) Resolve(ctx huma.Context) []error {
	i.Request, _ = humago.Unwrap(ctx)
	return nil
}

// Here I'm struggling to get the Input value
type Blended[I any] struct {
	RequestAware
	Input I
}

// The wrapper function
func Wrapper[I, O any](
	handler func(context.Context, *I) (*O, error),
) func(context.Context, *Blended[I]) (*O, error) {
	return func(ctx context.Context, t *Blended[I]) (*O, error) {
		doSomethingWithRequest(t.Request)
		return handler(ctx, &t.Input)
	}
}
```

In the above example, I'm struggling to get the Input value. The type `[I]` in my first registered handler, is like this:
```go
struct {
    Body struct {
        V int `json:"v"`
    }
}
```

I can however successfully capture the `*http.Request`. Is it because that the `I` is not embedded in the `Blended` struct ? But there's no way to embed a generic type parameter currently AFAIK. 
Can someone please help me, guide me out of this storm, please! Thank you beforehand!

sinloss:
Closes #811 

Allow input struct to have fields tagged with `recursive:"true"` to enable a deeper params finding.

With this change, the input struct composing can now incorperate generic type parameters which are normally unable to get embedded. 
And with limited knowledge, I would conservatively say this change does not break existing behaviours.

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -439,7 +439,7 @@ func _findInType[T comparable](t reflect.Type, path []int, result *findResult[T]
 					result.Paths = append(result.Paths, findResultPath[T]{fi, v})
 				}
 			}
-			if f.Anonymous || recurseFields || deref(f.Type).Kind() != reflect.Struct {
+			if f.Anonymous || recurseFields || deref(f.Type).Kind() != reflect.Struct || f.Tag.Get("recursive") == "true" {
 				// Always process embedded structs and named fields which are not
 				// structs. If `recurseFields` is true then we also process named
 				// struct fields recursively.


wolveix:
Hey @sinloss! Thanks for this PR :) I'm currently exploring setting `recursive: true` for both `findParams` and `findHeaders`, which would resolve this issue. Kindly follow the PR here: #943

PR #819, opened 2025-05-08
fix: override group modifier

fourcels:
```go
	grp := huma.NewGroup(api, "/users")
	grp.UseSimpleModifier(func(o *huma.Operation) {
		o.Summary = "User Summary"
	})

	// Register the router here
	huma.Get(grp, "/current", h.GetCurrent, func(o *huma.Operation) {
		o.Summary = "Current User"
	})

```

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -2039,8 +2039,14 @@ func convenience[I, O any](api API, method, path string, handler func(context.Co
 		Path:        path,
 		Metadata:    map[string]any{},
 	}
-	for _, oh := range operationHandlers {
-		oh(&operation)
+	if grp, ok := api.(*Group); ok {
+		for _, oh := range operationHandlers {
+			grp.UseSimpleModifier(oh)
+		}
+	} else {
+		for _, oh := range operationHandlers {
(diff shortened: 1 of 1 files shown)


huma.go
```diff
@@ -2039,8 +2039,14 @@ func convenience[I, O any](api API, method, path string, handler func(context.Co
 		Path:        path,
 		Metadata:    map[string]any{},
 	}
-	for _, oh := range operationHandlers {
-		oh(&operation)
+	if grp, ok := api.(*Group); ok {
+		for _, oh := range operationHandlers {
+			grp.UseSimpleModifier(oh)
```
danielgtaylor, line 2044: Doesn't this modify the group for all the other operations that may get registered on it as well? I'm not sure this is the correct behavior.

What bug are you trying to fix with this?

scriptforfivem:
This is a well-structured codebase. I appreciate the clean separation of concerns.

huma.go
fourcels: router modifier  should  overrite  group modifier

PR #824, opened 2025-05-18
Fix: Support for nested pointer struct fields in CLI options

tomMoulard:
## Description

This PR adds support for nested options in CLI configurations, particularly
when using pointer-to-struct fields. It addresses an issue where the CLI would
panic with `panic: Unsupported option type: ptr` or `panic: Unsupported option type: struct`
when attempting to set values on fields within a nil pointer struct.

## Example Use Case (my use case).

I was trying to implement nested configuration options with pointers for a clean optional field pattern:

```go
type DatabaseConfig struct {
    Host     string `doc:"Database host"`
    Port     int    `doc:"Database port" default:"5432"`
    Username string `doc:"Database username"`
}

type AppConfig struct {
    Debug bool            `doc:"Enable debug mode"`
    DB    *DatabaseConfig `doc:"Database configuration"` // Here both ptr or direct would have been acceptable.
}
```

When attempting to use this pattern with CLI flags like
`--db.host localhost --db.port 5432`,
I encountered a panic error: `panic: Unsupported option type: struct`.
The fix in this PR allows for properly handling these nested pointer-to-struct
configurations, so users can effectively organize related options.

## Changes

- Improved handling of nested fields by properly initializing nil pointers
  before setting values
- Enhanced error handling with more descriptive messages and proper error
  wrapping

I tried to split every major change in its own commit, btw

## Testing
Added a new test case TestCLINestedOptions that verifies both direct struct
fields and pointer-to-struct fields work correctly with CLI arguments and
environment variables.

Oh ! And thanks for your awesome project 💯

### DIFF
--- a/humacli/humacli.go
+++ b/humacli/humacli.go
@@ -156,8 +156,7 @@ func (c *cli[O]) OnStop(fn func()) {
 	c.stop = fn
 }
 
-func (c *cli[O]) setupOptions(t reflect.Type, path []int) {
-	var err error
+func (c *cli[O]) setupOptions(t reflect.Type, path []int) error {
 	flags := c.root.PersistentFlags()
 	for i := 0; i < t.NumField(); i++ {
 		field := t.Field(i)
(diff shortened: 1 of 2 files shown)


tomMoulard:
Sorry, I got bated by my bad test 😢

danielgtaylor (approved):
Thank you!

PR #837, opened 2025-06-09
feat: add humagin.NewContext

bbedward:
Ran into difficulties testing a middleware that used `humagin.Unwrap` Since `ginCtx` is not exported, it's not easy to create a huma.Context compatible with humagin.Unwrap.

Adding `NewContext` like some of the other adapters seems logical


danielgtaylor (approved):
Thanks! :+1:

PR #872, opened 2025-07-29
feat: use openapi.json by default

fourcels:
`openapi.json` can directly view in browser


danielgtaylor:
@fourcels why do you want this change? The doc generator supports YAML just fine.

wolveix:
Hey @fourcels! I'm going to close this as I'm not sure there's any reason to switch from YAML as the default, particularly considering that both are served _anyway_ and YAML is the standard for OpenAPI specifications.

Issue #868, opened 2025-07-23
Add `format: uuid` in the openapi spec generation for UUID strings


NicoleStrel:
Hello! 

Looks like `uuid` string validation is [supported in huma](https://github.com/danielgtaylor/huma/blob/badf024b0795f089518475622985ca899c98e41e/docs/docs/features/request-validation.md?plain=1#L146), however when the openapi spec is generated it is not mentioned as a validation requirement for the string with `format: uuid`. Heres the openapi spec for `uuid`: https://spec.openapis.org/registry/format/uuid.html

Can we fix openapi generation for string formats (uuid, and any others)?

leonklingele:
ftr: we might soon get an `uuid` package in the stdlib! 🥳 https://github.com/golang/go/issues/62026

philon123:
here is a working example for generating UUID fields in the spec like:
```yaml
CompanyID:
  type: string
  format: uuid
```

Excerpt from my code:
```go
// uuidSchema tells Huma to generate {type: string, format: uuid} for uuid.UUID fields.
type uuidSchema struct{}

func (uuidSchema) Schema(r huma.Registry) *huma.Schema {
	return &huma.Schema{Type: huma.TypeString, Format: "uuid"}
}

func getHumaConfig() huma.Config {
	// Register custom formats for fields that just appear as string in regular OpenAPI specs
	registry := huma.NewMapRegistry("#/components/schemas/", huma.DefaultSchemaNamer)
	registry.RegisterTypeAlias(reflect.TypeFor[uuid.UUID](), reflect.TypeFor[uuidSchema]())

	return huma.Config{
		OpenAPI: &huma.OpenAPI{
			OpenAPI: "3.0.4", // openapi-codegen doesn't support 3.1.0 yet
			Info: &huma.Info{
				Title:   "My App",
				Version: "0.1.0",
			},
			Components: &huma.Components{
				Schemas: registry,
				SecuritySchemes: map[string]*huma.SecurityScheme{
					"bearer": {
						Type:         "http",
						Scheme:       "bearer",
						BearerFormat: "JWT",
					},
				},
			},
		},
		OpenAPIPath:   "/openapi",
		DocsPath:      "/docs",
		Formats:       huma.DefaultFormats,
		DefaultFormat: "application/json",
	}
}
```

PR #875, opened 2025-07-31
Added uuid to have a default `format: uuid`

NicoleStrel:
Addresses https://github.com/danielgtaylor/huma/issues/868

Since `uuid` is a standard format, I added `format: uuid` so that you do not need to define a schema for this common type. 

Let me know if other [built-in string types](https://github.com/danielgtaylor/huma/blob/badf024b0795f089518475622985ca899c98e41e/docs/docs/features/request-validation.md?plain=1#L128-L146) need default formats

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/danielgtaylor/huma/v2/validation"
+	"github.com/google/uuid"
 )
 
 // ErrSchemaInvalid is sent when there is a problem building the schema.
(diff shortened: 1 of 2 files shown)


schema.go
```diff
@@ -45,6 +46,7 @@ var (
 	ipAddrType     = reflect.TypeOf(netip.Addr{})
 	urlType        = reflect.TypeOf(url.URL{})
 	rawMessageType = reflect.TypeOf(json.RawMessage{})
+	uuidType       = reflect.TypeOf(uuid.UUID{})
```
victoraugustolls: IMO users should be able to inform which uuid library they are using, as a google library might not be their first option. @NicoleStrel What do you think?

wolveix:
Hi @NicoleStrel! I'm going to close this for now due to it adding an additional dependency.

PR #876, opened 2025-08-01
Added exhaustive error code wrappers

rhodeon:
I was working on validating an image payload and noticed there was no wrapper for code 413 (Request entity too large), so I added one for that and all other error codes defined in the standard library (including error 418 XD).


wolveix:
Hah, this is great, and something I've reached for in the past! Thanks so much!

PR #891, opened 2025-09-12
omit patternDescription from yaml schema

7opf:
[JSON Schema Validation](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-01) defines the [pattern](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-01#name-pattern) keyword, but does not define any patternDescription keyword. This keyword seems to be used only by Huma.

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -112,7 +112,7 @@ type Schema struct {
 	MinLength            *int                `yaml:"minLength,omitempty"`
 	MaxLength            *int                `yaml:"maxLength,omitempty"`
 	Pattern              string              `yaml:"pattern,omitempty"`
-	PatternDescription   string              `yaml:"patternDescription,omitempty"`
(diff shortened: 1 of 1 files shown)


7opf:
It's not clear to me whether this is the correct fix. I just saw that the `Nullable` field is omitted from yaml serialisation, possibly for the same reason.

wolveix:
Thanks for opening this @7opf! Looking back, this was implemented in #323 by @purin-dev, and @danielgtaylor directly added the line to our JSON marshaler. I can't tell if it's intentional for it to be within the schema or not; however, I do agree that I'm not sure it should be (at best, it should be marshaled as `x-patternDescription` as that's the standard convention for OpenAPI extensions). Will await further clarity from the others :)

PR #905, opened 2025-10-12
fix: use "http" URL scheme only if no TLS is used

leonklingele:
This fixes an issue where the "http" schema was incorrectly used for the "$schema" value when https://localhost or https://127.0.0.1 was requested.

### DIFF
--- a/transforms.go
+++ b/transforms.go
@@ -175,7 +175,7 @@ func (t *SchemaLinkTransformer) Transform(ctx Context, status string, v any) (an
 
 	// Set the `$schema` field.
 	buf := bufPool.Get().(*bytes.Buffer)
-	if len(host) >= 9 && (host[:9] == "localhost" || host[:9] == "127.0.0.1") {
+	if ctx.TLS() == nil && len(host) >= 9 && (host[:9] == "localhost" || host[:9] == "127.0.0.1") {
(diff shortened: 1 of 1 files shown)


wolveix:
Thank you!

leonklingele:
@danielgtaylor is back in business! 🥳🥳🥳🥳🥳

leonklingele:
Oh wait! @wolveix, are you a new maintainer of `huma`?

wolveix:
I am! I volunteered to help lighten the load 🙏

leonklingele:
That's so nice to see, @wolveix ❤️ Not trying to discredit @danielgtaylor's absolutely amazing work on `huma` -- does it make sense to create some kind of `humago` GitHub org and move the repo(s)? I'd love to be more involved in the future, too, and would be very interested in helping maintain huma. Really trying to make this the # 1 HTTP framework. So many opportunities to make `huma` even more fun to use. :)

wolveix:
Couldn't agree more with you! I've been pushing Huma anywhere and everywhere for the past 2 years.

Regarding the org, I'll leave that entirely to @danielgtaylor. I think it's fine either way I think. Long-term an org would probably be better, but I'm not sure it's necessary right now :)

PR #916, opened 2025-11-17
chore(api,docs): update unpkg libraries, specify SRI hash, add strict CSP

leonklingele:
See the individual commits for details.

### DIFF
--- a/api.go
+++ b/api.go
@@ -595,9 +595,10 @@ func (a *api) registerDocsRoute() {
 		body = []byte(`<!doctype html>
 <html lang="en">
   <head>
-    <title>` + title + `</title>
     <meta charset="utf-8">
+    <meta name="referrer" content="no-referrer">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>` + title + `</title>
   </head>
   <body>
     <script id="api-reference" data-url="` + openAPIPath + `.json"></script>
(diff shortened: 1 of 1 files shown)


wolveix:
Any reason not to share CSPs between the docs renderers?

leonklingele:
> Any reason not to share CSPs between the docs renderers?

@wolveix they all need to have their own distinct CSP where only some directives are shared, while others depend on the UI framework being used.

leonklingele:
@wolveix can you have a look? :) This would be very nice to have included in an upcoming release so we can finally get rid ouf our custom doc handler and instead use one of Huma's default handlers 😃

leonklingele:
ftr: Chrome just fixed a zero day use-after-free in its CSS handling: https://chromereleases.googleblog.com/2026/02/stable-channel-update-for-desktop_13.html, this change will prevent unpkg.com from exploiting it (which will most probably _not_ happen, but still.)

wolveix:
Sorry, was just going through the other issues :)

Definitely keen to merge this before I create a new release; however, I'm not a fan of the current implementation (more on my side than yours). Reckon we could cut down on the duplicate code here, without making it an abstracted mess?

Not a blocker by any means, but if you have ideas now _before_ we merge, it'd be awesome to get them implemented 🙏🏻

leonklingele:
@wolveix updated the docs to refer to the default docs implementation in `api.go`. This will remove all code duplications. ptal :)

wolveix:
@leonklingele nicely done! One minor comment 🙏🏻

leonklingele:
@wolveix updated, ptal, again :)

PR #924, opened 2025-11-25
fix: failed to parse time wrapper, close #844

### LINKED ISSUE #844: OptionalParam with time.Time giving 422 Unprocessable Entity
I am using the OptionalParam struct from the Request Inputs docs (https://huma.rocks/features/request-inputs/#custom-wrapper-types), see the definition below. I then want to use this for optional query params of type time.Time, but this results in the handler giving a response of 422 Unprocessable Entity with a message of "invalid date/time for format". If neither timeFrom nor timeto are passed in the actual request query, then everything works as expected.

Is there something missing in the `OptionalParam` implementation that causes it to not support time.Time?

Or are there any other methods of implementing optional parameters that I should rather use?

request: "http://localhost:8080/endpoint?timeFrom=2025-01-01T12:00:00Z"

input to handler: 
```go
type GetByFilterRequestDto struct {
	groupId OptionalParam[uuid.UUID]            `query:"groupId" required:"false" format:"uuid"`
	TimeFrom       OptionalParam[time.Time]   `query:"timeFrom" required:"false"`
	TimeTo         OptionalParam[time.Time]      `query:"timeTo" required:"false"`
}
```

Definition of `OptionalParam`:
```go
type OptionalParam[T any] struct {
	Value T
	IsSet bool
}

// Define schema to use wrapped type
func (o OptionalParam[T]) Schema(r huma.Registry) *huma.Schema {
	return huma.SchemaFromType(r, reflect.TypeOf(o.Value))
}

// Expose wrapped value to receive parsed value from Huma
// MUST have pointer receiver
func (o *OptionalParam[T]) Receiver() reflect.Value {
	return reflect.ValueOf(o).Elem().Field(0)
}

// React to request param being parsed to update internal state
// MUST have pointer receiver
func (o *OptionalParam[T]) OnParamSet(isSet bool, parsed any) {
	o.IsSet = isSet
}
```

zhenzou:


wolveix:
Thank you!

Issue #910, opened 2025-11-04
Duplicate Operation IDs should not be allowed


rhodeon:
In both the [OpenAPI specification](https://spec.openapis.org/oas/v3.1.0#fixed-fields-7) and [Huma's comments](https://github.com/danielgtaylor/huma/blob/689a87da011dd155d23b8324968f9f786fef5796/openapi.go#L702), the `OperationID` is described to be unique:
> Unique string used to identify the operation. The id MUST  be unique among all operations described in the API. The operationId value is case-sensitive. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.

But this behaviour isn't enforced by Huma.
In the snippet below, the `list-users` OperationID is shared by both operations.
```go
huma.Register(
		api,
		huma.Operation{
			Method:      http.MethodGet,
			Path:        "/users",
			Summary:     "List users",
			OperationID: "list-users",
		},
		func(ctx context.Context, _ *struct{}) (*struct{}, error) {
			return nil, nil
		},
	)

	huma.Register(
		api,
		huma.Operation{
			Method:      http.MethodDelete,
			Path:        "/users",
			Summary:     "Delete users",
			OperationID: "list-users",
		},
		func(ctx context.Context, _ *struct{}) (*struct{}, error) {
			return nil, nil
		},
	)
```

Huma accepts this without errors and that results in a documentation page where the "List users" page cannot be selected as both are seen as the same endpoint:

<img width="1018" height="272" alt="Image" src="https://github.com/user-attachments/assets/f23587c8-d56f-49e5-9c69-5ea6bfdeef4b" />

To be inline with the specification and prevent mistakes, it would be good for Huma to reject duplicate operation IDs.
I think it would be better to have this behaviour as the default rather than an option as having duplicate IDs is a bug in any case, even if such a change could break some pre-existing code.

wolveix:
Hey @rhodeon, thanks for opening this issue and identifying this bug! I've opened a PR to resolve this; however, I don't want to merge it without @danielgtaylor's review given that I actually agree with you that the default should be denying operation ID duplicates, which makes it a breaking change. My PR implements it as a configurable parameter to avoid it being a breaking change, but I'd like Daniel's opinion before proceeding with it. Perhaps he'd agree that it should also panic by default :)

wolveix:
Hey @rhodeon! This is now the default behaviour :) This hasn't been included in a release yet, but it will be in the next release.

PR #925, opened 2025-11-25
Implement Registry and Query Validation Options

### LINKED ISSUE #524: Enable global additional properties
Is there an easy way to enable the `additionalProperties` on every request object through the config?

### LINKED ISSUE #660: Feature request: strict validation for query parameters
According to [the documentation](https://huma.rocks/features/request-validation/#strict-vs-loose-field-validation) Huma is strict about unexpected fields:

> By default, Huma is strict about which fields are allowed in an object, making use of the additionalProperties: false JSON Schema setting. This means if a client sends a field that is not defined in the schema, the request will be rejected with an error. This can help to prevent typos and other issues and is recommended for most APIs.

However, this does not currently apply to query parameters. Any additional query parameters are silently discarded. For example, here is the example from the tutorial where I changed `name` to be a query parameter:
```go
package main

import (
	"context"
	"fmt"
	"net/http"

	"github.com/go-chi/chi/v5"

	_ "github.com/danielgtaylor/huma/v2/formats/cbor"
)

// GreetingOutput represents the greeting operation response.
type GreetingOutput struct {
	Body struct {
	}
}

func main() {
	// Create a new router & API
	router := chi.NewMux()

	// Register GET /greeting/{name} handler.
	huma.Get(api, "/greeting", func(ctx context.Context, input *struct {
		Name string `query:"name" maxLength:"30" example:"world" doc:"Name to greet"`
	}) (*GreetingOutput, error) {
		resp := &GreetingOutput{}
		resp.Body.Message = fmt.Sprintf("Hello, %s!", input.Name)
		return resp, nil
	})

	// Start the server!
	http.ListenAndServe("127.0.0.1:8888", router)
}

```

If the API is requested with a misspelled query parameter (`firstname` instead of `name`), then no error is returned:

```bash
restish :8888/greeting\?firstname=world
HTTP/1.1 200 OK
Content-Length: 81
Content-Type: application/cbor
Date: Wed, 27 Nov 2024 08:58:42 GMT
Link: </schemas/GreetingOutputBody.json>; rel="describedBy"

{
  $schema: "http://localhost:8888/schemas/GreetingOutputBody.json"
  message: "Hello, !"
}
```

It would be nice to have the option to be strict about unexpected query parameters and return an error in these cases.

wolveix:
This PR:
- Adds `Config()` func to the `API` interface definition
  - This was necessary for the `Register` func to retrieve the new `RejectUnknownQueryParameters` config value (closes #660)
- Adds `Config()` func to the `Registry` interface definition, and a new `RegistryConfig` type
  - This was necessary for various internal funcs to retrieve the new `AllowAdditionalPropertiesByDefault` and `FieldsOptionalByDefault` config options (closes #524)

I'm not 100% confident in modifying the `API` and `Registry` interfaces, which is why I've held this PR for a few months to think about it. I'm struggling to think of a better solution, and this implementation doesn't introduce any significant downsides other than increasing the size of the interface definitions in a very common way.

### DIFF
--- a/docs/docs/features/request-validation.md
+++ b/docs/docs/features/request-validation.md
@@ -24,6 +24,7 @@ The standard `json` tag is supported and can be used to rename a field. Any fiel
 Fields being optional/required is determined automatically but can be overridden as needed using the logic below:
 
 1. Start with all fields required, **except for cookie, header, and query parameters which are optional by default**.
+    - If `huma.Config.FieldsOptionalByDefault` is set to `true`, then all fields are optional by default.
 2. If a field has `omitempty`, it is optional.
 3. If a field has `omitzero`, it is optional.
 4. If a field has `required:"false"`, it is optional.
@@ -35,7 +36,7 @@ Pointers have no effect on optional/required. The same rules apply regardless of
 
 ```go
 type MyStruct struct {
-    // The following are all required.
+    // The following are all required by default.
     Required1 string  `json:"required1"`
     Required2 *string `json:"required2"`
     Required3 string  `json:"required3,omitempty" required:"true"`
@@ -48,6 +49,15 @@ type MyStruct struct {
     Optional5 string  `json:"optional5" required:"false"`
 }
 ```
+
+If you prefer all fields to be optional by default, you can set `FieldsOptionalByDefault` in the `huma.Config`:
+
+```go
+config := huma.DefaultConfig("My API", "1.0.0")
+config.FieldsOptionalByDefault = true
+```
+
+With this setting, `Required1` and `Required2` in the example above would become optional. `Required3` would remain required.
 
 !!! info "Note"
 
@@ -201,6 +211,29 @@ type PartialInput struct {
 !!! info "Note"
 
     The use of `struct{}` is optional but efficient. It is used to avoid allocating memory for the dummy field as an empty object requires no space.
+
+## Unknown Query Parameters
+
+By default, unknown query parameters are ignored. You can set `RejectUnknownQueryParameters` to `true` in your `huma.Config` or for a specific `huma.Operation` to change this.
+
+```go title="config.go"
+config := huma.DefaultConfig("My API", "1.0.0")
+config.RejectUnknownQueryParameters = true
+```
+
+Or for a specific operation:
+
+```go title="operation.go"
+huma.Register(api, huma.Operation{
+    Method: http.MethodGet,
+    Path: "/test",
+    RejectUnknownQueryParameters: true,
+}, func(ctx context.Context, input *struct{}) (*struct{}, error) {
+    // ...
+})
+```
+
+When enabled, any query parameter sent by the client that is not defined in the input struct will result in a `422 Unprocessable Entity` error. This behavior is skipped if `SkipValidateParams` is set to `true` on the operation.
 
 ## Advanced Validation
 


wolveix:
Looking back on this, I think there's a lot of similar requests around #910. So I wonder if we should instead have a shared `ForceExplicitDefinitions` (or a better name) that would enforce this logic for everything.

danielgtaylor:
@wolveix is there a list of things that should be strict but aren't and can't easily be the new default? Query params are a good one. Operation IDs should be enforced to be unique. I'm trying to think of what else should be stricter.

wolveix:
@danielgtaylor I _thought_ there was another, though after a quick search I've not been able to find it. Since unique operation IDs will become the default (I'll rework my PR and merge it tomorrow), I guess it's just query params.

It'd be ideal if we could make it configurable from the config struct rather than the API struct. I couldn't see a way to do it when I opened this PR, though I'm much more intimately familiar with Huma's codebase now so I'd happily give it another look! That is, unless you have a better idea ;)

wolveix:
@leonklingele if you fancy getting stuck in, would you mind reviewing this? I'm very cautious about adding `.Config()` to the `Registry` interface, but I'm struggling to see any other path. I haven't created a Huma release since merging in the `FieldsOptionalByDefault` PR, for this exact reason.

I tried moving the fields over to the `OpenAPI` and `Operation` structs (allowing per-route overrides on top of global overrides), but this didn't really work because internal schema funcs still depended on the Registry struct.

I think this implementation is _fine_, and I've had colleagues review this PR too. But...I want to be sure :D

wolveix:
Thanks for reviewing @scottenock 🙏🏻 

@danielgtaylor would you mind taking a look at this? I'm very wary about adding `Config()` onto the `API` and `Registry` interfaces, but I also need to get this merged ASAP for some projects I'm working on. Would really appreciate your thoughts if you can spare the time 🙏🏻

wolveix:
An alternative could be:
```go
// ConfigProvider is an internal interface to get the configuration from an
// implementation of the API or Registry. This is used by the `huma` package
// to access settings without exposing them on the public interfaces.
type ConfigProvider[T any] interface {
    Config() T
}
```

But while that removes `Config()` from the interface definitions. It's still usable via:
```go
api.(*huma.api).Config()
```

huma.go
```diff
@@ -681,6 +703,60 @@ func Register[I, O any](api API, op Operation, handler func(context.Context, *I)
...
+				for _, prefix := range deepPrefixes {
+					if strings.HasPrefix(key, prefix) {
+						isDeep = true
+						break
+					}
+				}
+
+				if isDeep {
+					continue
+				}
```
leonklingele:
```suggestion
				isDeep := false
				for _, prefix := range deepPrefixes {
					if strings.HasPrefix(key, prefix) {
						// Is deep
						continue outer
					}
				}
```

leonklingele (commented):
Sorry for dropping in so late to the review!

Just two minor things I noticed we could improve, and a question regarding your suggestion in https://github.com/danielgtaylor/huma/pull/925#issuecomment-3910598665: How would you use the `ConfigProvider` interface, and where? 🤔

wolveix:
Thanks @leonklingele!

With `ConfigProvider`, you'd remove the `Config()` funcs from the `API` and `Registry` interfaces, instead calling them using generics. It would mean that despite the interface not requiring a `Config()` func, the concrete types could still implement them and be usable. I'm not a huge fan of the implementation, and no longer have the mock implementation locally.

The main thing I'm trying to decide is whether exposing `Config()` on both of those structs is acceptable or not. I kinda feel like yes, because it gives more flexibility to Huma consumers; however, it's at the cost of exposing internal config values (albeit as read-only). I've spent the last few months mentally chewing it, and this is the best implementation I've been able to come up with.

leonklingele:
> With ConfigProvider, you'd remove the Config() funcs from the API and Registry interfaces, instead calling them using generics.

Sorry, I can't really wrap my head around this 🙈 I'm confused by the `api.(*huma.api).Config()` you mentioned above. Can you share some code to make this more clear to me? (It's been a long day already! 😅)

wolveix:
@leonklingele I threw this together quickly to give an example: https://github.com/wolveix/huma/commit/31a24086c4256cd6f42e80b5e1f993a674535882

I added `GetConfig()` to give a bit more of an idea. Like I said, I'm not really a fan of this approach either, _but_ it does remove `Config()` from the interface definitions.

leonklingele:
@wolveix thanks, that makes it clear now! :)

This one is tricky. Personally, I'd prefer to not change the interfaces so we don't break existing implementations, and instead let consumers decide if they want to provide any custom config or not. There's not really a good reason to force them into satisfying an adjusted interface and returning some of Huma's default configs when they don't care about changing the default config at all.

On the other hand, I prefer to be as explicit as possible and not introduce any "magic" interfaces which change internal behavior 🤔

What are the downsides of deferring that decision for now and making the func optional first (i.e. going with https://github.com/wolveix/huma/commit/31a24086c4256cd6f42e80b5e1f993a674535882), and simply adapting the interface on a new major release if we deem it necessary?

wolveix:
I hear you! And I really appreciate you taking the time to review this :) 

I guess the downside of doing so would be that it'd be something we may want to deprecate in the future. I'm responding from my phone right now, but a common practice in Go is to accept interfaces but return concrete types...I wonder if we could revise our internal funcs to work off of the now-private (but maybe shouldn't be) `api` struct? Because if so, that completely resolves this issue, as funcs would then have access to the private vars, since they're defined within the same package.

I'll look at this tomorrow, perhaps my tired eyes are missing an obvious reason why that wouldn't work :) But if so, the generic `GetConfig` would seemingly be the way to go.

danielgtaylor:
@wolveix and others thanks for really digging into this one. It feels like we have three options for the implementation:

1. New global config options like `huma.AllowAdditionalPropertiesByDefault` and `huma.FieldsOptionalByDefault` that would mimic the existing ones like `huma.ValidateStrictCasing` and `huma.DefaultArrayNullable`.
2. The original proposal here, with an updated `API` interface and plumbing things through.
3. The proposal in https://github.com/wolveix/huma/commit/31a24086c4256cd6f42e80b5e1f993a674535882

I took some time to look it all over. Any reason to avoid option 1? The main one I can think of is that global state is a bit of a smell and prevents multiple API instances in the same process from having different defaults. But... we do have some existing precedent for it.

As for the other two, I prefer option 3 to leave the existing interfaces alone if we can. As is mentioned above that leaves it open for us to still modify interfaces at a later date if desired, so we don't back ourselves into a corner, and this is similar to what the Go standard library does as well, e.g. [`http.ResponseController`](https://pkg.go.dev/net/http#ResponseController) wraps the response and uses optional interfaces that can be implemented and came later.

Again thanks for the awesome work on this! I really appreciate it.

wolveix:
Thanks so much for your review @danielgtaylor!

I've taken the generics implementation, and have made both `GetConfig` and `ConfigProvider` into private declarations. I'm finally happy with this implementation 🤩

Issue #833, opened 2025-05-28
Custom Content-Type negotiation issues


k0ality:
I want to use custom Content-Type in Format, DefaultFormat, Input Body, Output. It's pure json, but I'm restricted from "application/json", "json" types.

When I add my custom type in Input Body I face HTTP 415 when I send no "Content-Type" header in request or a crash if I add "Content-Type" header.

Here's a runnable playground example: https://go.dev/play/p/FWy4_38Rh9z

giraffesyo:
@k0ality I was able to achieve this in my app:

```go
package routes

import (
	"context"
	"net/http"

	"github.com/danielgtaylor/huma/v2"
	"github.com/parallelworks/core/packages/go/ingress/internal/auth"
)

// WhoAmIOutput represents the whoami operation response.
type WhoAmIOutput struct {
	ContentType string `header:"Content-Type"`
	Body        []byte `example:"user:jakethayne" doc:"The subject of the request, typically a user or the subject of the token."`
}

func RegisterRoutes(api huma.API) {
	huma.Register(api, huma.Operation{
		OperationID: "whoami",
		Method:      http.MethodGet,
		Path:        "/api/auth/whoami",
		Summary:     "Identify the current user",
		Description: "Returns the currently authenticated subject.",
		Tags:        []string{"Authentication"},
	}, func(ctx context.Context, input *struct{}) (*WhoAmIOutput, error) {
		resp := &WhoAmIOutput{}

		// get the subject from the context
		subject := auth.GetSubject(ctx)

		resp.Body = []byte(subject)
		resp.ContentType = "text/plain"
		return resp, nil
	})

}

```

richard-hunter-hpe:
Using a named type for the body itself, and putting the tag there is what ended up working for me:

```go
type HelloInputReqBody struct {
	Name string `json:"name" minLength:"1" maxLength:"256" contentType:"test-ct-custom"`
}

type HelloInput struct {
	Body HelloInputReqBody
}
```

Working example: https://go.dev/play/p/qMJZU0P67_0


Putting the tag on the body panics at runtime: https://go.dev/play/p/vCZluWyY3KJ

richard-hunter-hpe:
@danielgtaylor will merging https://github.com/danielgtaylor/huma/pull/926 fix this? I've tested the change against a custom `contentType` and it seems to work.

wolveix:
[...]

PR #926, opened 2025-11-27
Fix: Support non-JSON content types for request body validation

roidelapluie:
Previously, Huma was hardcoded to look for the validation schema in RequestBody.Content["application/json"]. This caused validation to fail when operations only specified other content types like application/cbor or application/x-www-form-urlencoded.

This commit makes the schema lookup more flexible:
1. First try application/json for backwards compatibility
2. If not found, fall back to the first available content type with a schema

This allows operations to use contentType tags to specify non-JSON request bodies while still benefiting from Huma's validation.

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -1187,7 +1187,7 @@ func processInputType(inputType reflect.Type, op *Operation, registry Registry) 
 		} else {
 			// Fall back to first available content type with a schema.
 			for _, mediaType := range op.RequestBody.Content {
-				if mediaType.Schema != nil {
+				if mediaType.Schema != nil && mediaType.Schema.Type != "string" && mediaType.Schema.Format != "binary" {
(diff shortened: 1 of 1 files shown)


richard-hunter-hpe:
@roidelapluie will this address https://github.com/danielgtaylor/huma/issues/833? I've been trying to use `application/merge-patch+json` Content Type on an input body, and while it works in the spec generation, get runtime errors during input validation due to a nil schema reference. With this change things seem to work. It would be great to get this bug fixed.

wolveix:
Hey @roidelapluie! This is an awesome change, thanks so much for contributing!

wolveix:
@richard-hunter-hpe no, this doesn't resolve that issue. But I have a fix locally for it, and will push it shortly :)

PR #927, opened 2025-12-03
Fix code samples in the response output documentation

prnvbn:
a very small PR fixing code samples in the response output documentation


wolveix:
Thanks so much!

PR #928, opened 2025-12-04
fix: use Body() instead of BodyRaw() in fiber to auto decompress the req body

akkuman:
BodyRaw() does not automatically recognize the value in Content-Encoding to decompress the request body, and returns unexpected values in the case of request compression


wolveix:
This looks consistent with our other adapters, thanks :)

PR #932, opened 2025-12-27
fix: prevent panic when text value sent to FormFile field

### LINKED ISSUE #912: Got panic when sending a string in file field in a form
I got panic, and its hard to figuring out earlier:
https://github.com/danielgtaylor/huma/blob/689a87da011dd155d23b8324968f9f786fef5796/huma.go#L1567

IMO, it should return validation error if the file field mistakenly send as a string value, not panic

alexisvisco:
Skip FormFile and []FormFile fields in formValueParser to prevent panic when multipart form data contains text values for fields expecting file uploads.

Previously, sending a text value to a FormFile field would cause parseInto() to panic with 'unsupported param type huma.FormFile'. Now it properly returns a validation error 'File required'.

Partially fixes #912

### DIFF
--- a/huma_test.go
+++ b/huma_test.go
@@ -1413,7 +1413,7 @@ Hello, World!
 					Path:   "/upload",
 				}, func(ctx context.Context, input *struct {
 					RawBody huma.MultipartFormFiles[struct {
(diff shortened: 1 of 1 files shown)


wolveix:
Awesome, thanks so much!

Issue #870, opened 2025-07-24
Enabling recursive process structs for patameters


birukbelay:
On huma parameters are not checked recursively, on structs which limits us to only embedding structs for queries
on humas code on huma/huma.go line 213:  it sets recurseFields to false,
, false, "Body")

but because of that it prevents use of nested structs  specially for such cases
func (uh *IGenericController[T, C, U, F, Q]) OffsetPaginated(ctx context.Context, filter *struct {
	Filter F
	Query  Q
	dtos.PaginationInput
}) (*dtos.HumaResponse[dtos.PResp[[]T]], error) {
	
	resp, err := DbFetchManyWithOffset[T](uh.GormConn, ctx, filter.Filter, filter.PaginationInput)
	return dtos.PHumaReturn(resp, err)
}

because **F and Q** are Generic type i could not just embed them with out **Filter & Query** fields,
so only by setting the value on huma/huma.go:213 we can fix this problem

repelliuss:
I would also like to know why we can't use struct embedded types or how to use them for request types.

For example, I would like to do,

```go
type pagination struct {
	Limit  int32 `query:"limit" default:"20" minimum:"1" maximum:"100"`
	Offset int32 `query:"offset" default:"0" minimum:"0"`
}

type email struct {
	Email string `query:"Email" format:"email" example:"user@example.com"`
}

func (s *Server) ListUsers(ctx context.Context, input *struct{pagination; email}) (*response[userArray], error) {
}
```

effectively

PR #943, opened 2026-01-17
Improve and Fix Header Parsing

### LINKED ISSUE #301: Nested header on input
I see that it is a common pattern with `Huma` that the body, both in the input and output, is to be created inside the nested struct field `Body`. Is it possible to do the same with headers and path/query parameters? Something like:

```go
type MyInput struct {
    Headers struct {
        UserIP string `header:"user-ip"`
    }
    Body struct {
        Name string `json:"name"`
    }
}
```

### LINKED ISSUE #811: How to create a wrapper func that also requires the *http.Request object?

wolveix:
#944 was merged as a successor to the original intention of this PR.

---

This PR enables recursion for `findInType` for `findHeaders` and `findParams`. I'm still exploring the implications of doing this, though my initial findings don't turn up anything negative. The main purpose of this is to allow named fields in input and output structs, which have been requested features for quite some time.

This PR:
- Fixes #301 by permitting headers being explicitly set from a struct
- Fixes #811 and #870


wolveix:
@leonklingele if you get a chance, I'd love your thoughts on this :)

leonklingele:
@wolveix sorry for the very late reply. This lgtm! Let's get this merged?

leonklingele:
Hey @wolveix, I'm now in a situation where I'd benefit from this change 😄 Is there anything still missing here?

wolveix:
I kinda wanted @danielgtaylor's feedback on it before merging 😅

wolveix:
@leonklingele I updated this PR, would appreciate your review :)

huma.go
```diff
@@ -306,6 +306,27 @@ type headerInfo struct {
...
 
+// isPromotedField reports whether the field identified by the index path is
+// reachable from the root type purely through embedded (anonymous) structs, so
+// Go promotes it as if it were declared at the top level. Such fields keep the
+// legacy behavior of being auto-named as a header from the field name; named
+// (non-embedded) nested struct fields must opt in with an explicit `header`
+// tag instead.
+func isPromotedField(root reflect.Type, path []int) bool {
+	t := baseType(root)
+	for _, idx := range path[:len(path)-1] {
```
leonklingele, line 317: Can `path` ever be empty? 🤔

huma_test.go
```diff
@@ -2187,19 +2259,42 @@ Content-Type: text/plain
...
+				type HiddenHeadersSliceElem struct {
+					HiddenWithTag    string `header:"X-Hidden-With-Tag-Slice"`
+					HiddenWithoutTag string // No header tag - nested (not promoted), so NOT set as a header.
 				}
 
 				type Resp struct {
 					*HiddenHeaders `hidden:"true"`
 
+					// Hidden slice field to exercise hidden-walk across slice -> elem.
+					HiddenSlice []HiddenHeadersSliceElem `hidden:"true"`
```
leonklingele, line 2296: Will this also hide nested structs?

```
type Auth struct {
    Token string `header:"X-Token"`
}

type Input struct {
    Auth Auth `hidden:"true"` // <-- I'd expect this to be hidden
}
```

huma.go
leonklingele, line 1: Does this really fix #811? 

`processInputType` still only detects a request body via a top-level lookup: 
https://github.com/danielgtaylor/huma/blob/f7f7e18d323197dd8180fd010a3ad20295219456/huma.go#L1433

So this shape still will not populate Input.Body:

```go
type Inner struct {
    Body struct {
        V int `json:"v"`
    }
}

type Blended struct {
    RequestAware
    Input Inner
}
```

(See https://github.com/danielgtaylor/huma/issues/811#issuecomment-2869662840)

huma.go
```diff
@@ -382,6 +417,65 @@ func (r *findResult[T]) Every(v reflect.Value, f func(reflect.Value, T)) {
...
+		}
+		current = current.Elem()
+	}
+
+	if current.Kind() == reflect.Invalid {
+		return false
+	}
+
+	set := false
+	switch current.Kind() {
```
leonklingele, line 448: What about arrays? 😅

```go
type Header struct {
    Value string `header:"X-Value"`
}

type Resp struct {
    Headers [1]Header
}
```

PR #952, opened 2026-02-02
Remove default value in for Type in ErrorModel

hiddenmarten:
Hi,

I'm not sure if this "fix" is really necessary, but I faced an issue using `huma` with `openapigenerator`.

The thing is `about:blank` is not suitable for the most of uri validators (in my case one included in R client), so my solution was to remove it from the result openapi spec, probably it would be helpful to remove or update it here.

Thanks for your project, it's great!

### DIFF
--- a/error.go
+++ b/error.go
@@ -69,7 +69,7 @@ func (e *ErrorDetail) ErrorDetail() *ErrorDetail {
 //	}
 type ErrorModel struct {
 	// Type is a URI to get more information about the error type.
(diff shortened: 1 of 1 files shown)


wolveix:
Hey @hiddenmarten! According to the RFC, this value _should_ default to `about:blank` when no explicit value is given: https://www.rfc-editor.org/rfc/rfc9457.html#section-3.1.1

I don't have particularly strong feelings either way. It's a little odd that `openapi-generator` doesn't support it.

PR #965, opened 2026-02-16
Export Gorilla Mux interface constructors

RangelReale:
Export interface constructors like the ones available in the stdlib Go wrapper.


wolveix:
Thank you!

PR #976, opened 2026-02-19
fix(group): ensure "Group" satisfies the configProvider interface

leonklingele:
Before this change, using Huma's route group feature would not propagate Huma's config to the group router and cause all routes attached to a group to use a default Huma config.

### DIFF
--- a/api.go
+++ b/api.go
@@ -313,6 +313,9 @@ type api struct {
 	middlewares  Middlewares
 }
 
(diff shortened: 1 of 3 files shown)


wolveix:
Great catch! The tests should include group in their name, since currently they look the same as the normal tests.

I like the interface implementation requirement, though there should be a comment with them so they're clear to everyone :)

leonklingele:
Updated, ptal :)

PR #998, opened 2026-03-12
Adds support for slice query parameters using ParamWrapper types

sharonjl:
Adds support for slice query parameters using `ParamWrapper` types (e.g., `[]OptionalParam[int]`). Previously,
  `parseSliceInto` only handled primitive slice element kinds. Struct elements implementing `ParamWrapper` were
  returned as `errUnparsable`.

  ## Changes

  - Added `parseArrStructElement` helper that unwraps each slice element's `ParamWrapper.Receiver()` and
  delegates per-element parsing to `parseInto`, supporting all receiver types including `time.Time`, `url.URL`,
  and `TextUnmarshaler`
  - Added `case reflect.Struct` to `parseSliceInto` switch for `ParamWrapper`-implementing struct elements
  - Threaded `ctx` and `paramFieldInfo` through `parseSliceInto` so the struct case can call `parseInto` directly
  - `ParamReactor.OnParamSet` is called per element when implemented, with the unwrapped parsed value matching
  scalar behavior

I am a little unsure of the reflect unwrapping but there doesn't seem to be another approach other than implementing TextMarshal/Unmarshal to support slice query params of type struct.

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -648,6 +648,35 @@ func parseArrElement[T any](values []string, parse func(string) (T, error)) ([]T
 	return result, nil
 }
 
+// parseArrStructElement parses a slice of string values into a slice of
+// ParamWrapper struct elements. It sets the wrapper slice on f and returns the
+// unwrapped parsed values for validation.
+func parseArrStructElement(ctx Context, f reflect.Value, values []string, p paramFieldInfo) (any, error) {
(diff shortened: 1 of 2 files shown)


wolveix:
Thanks for this @sharonjl! #1021 ended up superseding it, but I'd happily re-open if it missed something :)

PR #1002, opened 2026-03-25
Add support for exploded query params

Ivan-Roger:
This allows for free-form query params in the case where they cannot be described in the openapi docs.
They will be stored as a map of strings (keys) to list of strings (values).

Example:

URI: `/search?name=foo&status=active`

```JSON
{
    "/search": {
        "get": {
            "summary": "Search items",
            "description": "Search items",
            "operationId": "search",
            "parameters": [
                {
                    "explode": true,
                    "in": "query",
                    "name": "filters",
                    "schema": {
                        "additionalProperties": {
                            "items": {
                                "type": "string"
                            },
                            "type": [
                                "array",
                                "null"
                            ]
                        },
                        "type": "object"
                    }
                }
            ]
        }
    }
}
```

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -41,6 +41,7 @@ var paramReactorType = reflect.TypeFor[ParamReactor]()
 var paramWrapperType = reflect.TypeFor[ParamWrapper]()
 var stringType = reflect.TypeFor[string]()
 var stringSliceType = reflect.TypeFor[[]string]()
+var urlValuesType = reflect.TypeFor[url.Values]()
 
 // Store int to string status number conversions for efficiency.
 var statusStrings = map[int]string{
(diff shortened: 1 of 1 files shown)


wolveix:
Hey @Ivan-Roger, could you give an example of what issue this is resolving? A complete code sample always helps cut down on the time necessary to review PRs :)

Ivan-Roger:
Hey,
Sorry. I openned the PR in a hurry. I added a description.
I'm available if you need any clarifications

wolveix:
Thanks for the PR and the example, it makes the intent clear!

I think this is mostly achievable today via a `Resolver`, without a change to the core param path:

```go
type SearchInput struct {
    Filters url.Values `json:"-"`
}

func (i *SearchInput) Resolve(ctx huma.Context) []error {
    u := ctx.URL()
    i.Filters = u.Query()
    return nil
}
```

Given that, I'd lean away from adding it to the param parser. The main reason is that it introduces a bit of unexpected magic: a field tagged `query:"filters"` doesn't actually read a `filters` param, it captures the whole query string, so the generated spec ends up describing something clients don't really send. huma tries to keep the spec an accurate description of the wire format, so I'd rather document the resolver pattern above than special-case this.

One heads up on the implementation regardless: gating on `p.Explode` also matches ordinary exploded arrays like `[]string` with `query:"tags,explode"`, which then hit the new type check and panic at request time (the existing `query:"exploded,explode"` test starts returning 500).

Happy to discuss if you feel strongly about the ergonomics though, curious what you think.

PR #1004, opened 2026-04-01
feat: add Schema.Const

seanwu1105:
Add support for [`const`](https://json-schema.org/understanding-json-schema/reference/const) in `Schema`.


wolveix:
Hey @seanwu1105! Thanks for opening this PR :) I'm intrigued, I'm not sure I've ever seen a need for this before. Could you give a little more context on what your use-case looks like? From @leonklingele's reaction, it seems you're not the only one that would benefit from it :D

seanwu1105:
Hey @wolveix, the main use case for me is the enum with name and value: https://stackoverflow.com/questions/66465888/how-to-define-enum-mapping-in-openapi

wolveix:
Cool! Thanks so much!

PR #1006, opened 2026-04-01
fix: support json inline tag for embedding anonymous fields in schema

lsdch:
Patches #978 to support `json:",inline"` tags for embedding anonymous fields. 

Although a Go embedded field is implicitly inlined, and the presence of this tag on embedded fields is thus redundant, this is still valid syntax and should be supported.


wolveix:
Hey @lsdch! Would you mind reviewing Copilot's comments, please? :)

lsdch:
Hi @wolveix, I believe I did in 1debcd3 
Sorry, I should have added a comment to signal it 😅
Please let me know if the PR still needs fixing

schema.go
```diff
@@ -703,9 +704,18 @@ func getFields(typ reflect.Type, visited map[reflect.Type]struct{}) []fieldInfo
 			continue
 		}
 
-		if f.Anonymous && f.Tag.Get("json") == "" {
-			embedded = append(embedded, f)
-			continue
+		if f.Anonymous {
```
fredDJSonos: beware that `json/v2` has the true support for `inline` on non anonymous fields.

```go
type Foo struct {
  X Bar `json:",inline"`
}
```
A serialization of a `Foo` will just _inline_ the serialisation of a `Bar`.

So you better hoist the logic outside the `if Anonymous`

fredDJSonos: This would be a better fix (totally untested)
```go
		jsonTag := f.Tag.Get("json")
		parts := strings.Split(jsonTag, ",")

		if len(parts) > 0 && parts[0] == "" && slices.Contains(parts[1:], "inline") {
			embedded = append(embedded, f)
			continue
		}

		if f.Anonymous && (len(parts) == 0 || parts[0] == "") {
			embedded = append(embedded, f)
			continue
		}
```

and it avoid the nasty situation where the user wants his field called `inline`

lsdch:
Thanks @fredDJSonos for the review !
I used the patch you suggested, and added a test for this case.

fredDJSonos (approved):

schema.go
wolveix: The non-anonymous ,inline support assumes json/v2 marshaling, but huma defaults to encoding/json v1 (given that v2 is still experimental), which nests named fields instead of inlining them. That makes the generated schema diverge from the actual wire format. I've corrected this behaviour for the time being in a new commit: [`aeccddd` (this PR)](https://github.com/danielgtaylor/huma/pull/1006/commits/aeccddd5d9a2d45e6df6f07cb84fc2ff2af33179) :)

PR #1011, opened 2026-04-12
fix: enforce required=true for path parameters per OpenAPI spec

### LINKED ISSUE #1009: Path parameters with required:"false" struct tag omit required field from OpenAPI spec (regression since v2.37.0)
### Description

When a path parameter struct field has `required:"false"` in its tag, huma v2.37.0+ omits the `"required"` field entirely from the generated OpenAPI spec. Per the [OpenAPI 3.x specification](https://spec.openapis.org/oas/v3.1.0#parameter-object), path parameters **MUST** have `required: true`:

> If the parameter location is "path", this property is REQUIRED and its value MUST be true.

This is a regression from v2.35.0, where `required:"false"` on path parameters was silently ignored and `required: true` was always emitted.

### Reproduction

```go
package main

import (
	"context"
	"fmt"
	"net/http"

	"github.com/danielgtaylor/huma/v2/humatest"
)

func main() {
	_, api := humatest.New(nil, huma.DefaultConfig("Test", "1.0.0"))

		OperationID: "trigger-error",
		Method:      http.MethodPost,
		Path:        "/debug/error/{code}",
		Summary:     "Trigger error",
		Code int `path:"code" required:"false" default:"403"`
	}) (*struct{}, error) {
		return nil, nil
	})

	b, _ := api.OpenAPI().MarshalJSON()
	fmt.Println(string(b))
}
```

**Expected output** (parameter object):
```json
{
  "in": "path",
  "name": "code",
  "required": true,
  "schema": { "default": 403, "type": "integer" }
}
```

**Actual output** (v2.37.3):
```json
{
  "in": "path",
  "name": "code",
  "schema": { "default": 403, "type": "integer" }
}
```

The `"required"` field is missing entirely. This causes downstream OpenAPI consumers (in our case, [orval](https://orval.dev/)) to reject the spec as invalid.

### Root Cause

this probably s due to to interaction between two code paths in `huma.go`:

**Step 1** — `parseParamLocation()` correctly sets `Required = true` for path params ([huma.go:149](https://github.com/danielgtaylor/huma/blob/v2.37.3/huma.go#L149)):
```go
case f.Tag.Get("path") != "":
    pfi.Loc = "path"
    pfi.Name = f.Tag.Get("path")
    pfi.Required = true
```

**Step 2** — Back in `findParams()`, the `required` tag override runs unconditionally **after** step 1 ([huma.go:245-247](https://github.com/danielgtaylor/huma/blob/v2.37.3/huma.go#L245-L247)):
```go
if _, ok = f.Tag.Lookup("required"); ok {
    pfi.Required = boolTag(f, "required", false)
}
```

This overwrites `Required` back to `false` for any path parameter tagged `required:"false"`.

**Step 3** — `Param.MarshalJSON()` uses `omitEmpty` for the `Required` field ([openapi.go:561](https://github.com/danielgtaylor/huma/blob/v2.37.3/openapi.go#L561)):
```go
{"required", p.Required, omitEmpty},
```

Since `false` is the zero value for `bool`, it gets omitted entirely from the output.

### Why this changed

In v2.35.0, the `required` tag handling was one-directional — it could only promote a parameter to required:
```go
if r := f.Tag.Get("required"); r == "true" {
    pfi.Required = true
}
```

PR #925 (v2.37.0) changed this to bidirectional to support the new `FieldsOptionalByDefault` config option. This inadvertently allowed `required:"false"` to override the mandatory `required: true` for path parameters.

Notably, in #863, @wolveix explicitly stated:
> "according to the OpenAPI specification, the path parameter is a strict requirement. Furthermore, supporting optional path parameters would generate non-compliant specs for OpenAPI consumers"

So I believe the current behavior is unintentional.

### Environment

- huma version: v2.37.3 (regression introduced in v2.37.0)
- Go version: 1.24
- Last working version: v2.35.0

P.S. this might be a possible fix:

adding a guard after the tag override in `findParams()`... something like:

```go
if _, ok = f.Tag.Lookup("required"); ok {
    pfi.Required = boolTag(f, "required", false)
}

// Path parameters must always be required per OpenAPI spec
if pfi.Loc == "path" {
    pfi.Required = true
}
```
but there may be a more appropriate place to enforce this invariant.

Yanhu007:
Fixes #1009

## Problem

Since v2.37.0, path parameters with `required:"false"` struct tag omit the `"required"` field from the generated OpenAPI spec. Per the [OpenAPI 3.x specification](https://spec.openapis.org/oas/v3.1.0#parameter-object):

> If the parameter location is "path", this property is REQUIRED and its value MUST be true.

The initial `Required = true` set for path params (line 149) was being overridden by the generic `required` tag processing (lines 245-246).

## Fix

Add post-override enforcement that always sets `Required = true` for path parameters, regardless of the struct tag value. This restores the pre-v2.37.0 behavior where `required:"false"` on path parameters was silently ignored.

### DIFF
--- a/huma.go
+++ b/huma.go
@@ -246,6 +246,12 @@ func findParams(registry Registry, op *Operation, t reflect.Type) *findResult[*p
 			pfi.Required = boolTag(f, "required", false)
 		}
 
+		// Per OpenAPI 3.x spec, path parameters MUST always be required.
+		// Override any user-set `required:"false"` tag for path params.
+		if pfi.Loc == "path" {
+			pfi.Required = true
+		}
+
 		if pfi.Type == timeType {
 			timeFormat := time.RFC3339Nano
 			if pfi.Loc == "header" {
--- a/huma.go
+++ b/huma.go
@@ -241,7 +241,9 @@ func findParams(registry Registry, op *Operation, t reflect.Type) *findResult[*p
 		pfi := pl.pfi
 		pfi.Schema = SchemaFromField(registry, f, "")
 
-		// While discouraged, make it possible to make query/header params required.
+		// While discouraged, make it possible to override `required` for non-path
+		// params via the struct tag. Path params are always required per the
+		// OpenAPI 3.x spec and are forced back to true below.
 		if _, ok = f.Tag.Lookup("required"); ok {
 			pfi.Required = boolTag(f, "required", false)
 		}
--- a/huma_test.go
+++ b/huma_test.go
@@ -600,6 +600,28 @@ func TestFeatures(t *testing.T) {
 			},
 		},
 		{
+			Name: "path-param-required-overrides-tag",
+			Register: func(t *testing.T, api huma.API) {
+				huma.Register(api, huma.Operation{
+					Method: http.MethodGet,
+					Path:   "/items/{id}",
+				}, func(ctx context.Context, input *struct {
+					ID string `path:"id" required:"false"`
+				}) (*struct{}, error) {
+					return nil, nil
+				})
+
+				// Per OpenAPI 3.x spec, path parameters must always be required,
+				// even if the struct tag says otherwise. Regression test for #1009.
+				param := api.OpenAPI().Paths["/items/{id}"].Get.Parameters[0]
+				assert.Equal(t, "id", param.Name)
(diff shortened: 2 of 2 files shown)


wolveix:
Thanks @Yanhu007!

PR #1014, opened 2026-04-14
fix: improve SSE stream response handling and add fasthttp support (#888)

### LINKED ISSUE #888: SSE: deadline failed to flush
Hi,

I'm trying to use streaming response and sse huma with fiber and i'm experiencing failed to flush like this issue: https://github.com/danielgtaylor/huma/issues/491

zhouyusd:
Fix #888

### DIFF
--- a/sse/sse.go
+++ b/sse/sse.go
@@ -2,9 +2,11 @@
 package sse
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"reflect"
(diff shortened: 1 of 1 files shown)


wolveix:
Hey @zhouyusd! Thanks for taking the time to implement this. Unfortunately, we're strictly against importing excessive dependencies in the core Huma library unless they're absolutely necessary; with your implementation, all Huma users would have to import `fasthttp`. With this in mind, I created #1059 which introduces a small internal streaming hook. I referenced you directly in that PR.

Thanks again!

PR #1019, opened 2026-04-29
Propagate underlying context to GET/PUT requests in the autopatch handler

alixinne:
Using NewRequest defaults to context.Background(), which breaks context chaining when handling PATCH requests. The outer context may contain values such as per-request information, loggers, etc. which should be propagated to the GET/PUT requests to ensure there's no difference between calling those requests directly and through a PATCH.


wolveix:
Thank you!

Issue #488, opened 2024-06-24
Handling of custom types for List inputs of query params


mvrahden:
So currently huma supports list inputs for input parameters of `query` type.

However, the support seems to be limited to `[]string`-type only.
It would be handy to have the support of string-serializable custom types in order to not perform the type parsing and mapping manually.

One example:
```go
// LicensesListFilterInput represents the list filter to match licenses.
type LicensesListFilterInput struct {
	LicenseIDs []uuid.UUID `query:"ids" required:"false" example:"a60a4a7e-33bb-4337-a72c-d33aa195936e,5b50ae19-f590-4921-871d-2ab6005a7335,35ed2d12-50d7-4939-bd54-5146e6416797"`
	// ...
}
```

Is there an easy way to achieve this?

Supporting the `UnmarshalText(data []byte) error` interface would be sufficient to support a wide range of types.

wolveix:
This should be resolved by #1021, which will be included in our next release :)

PR #1021, opened 2026-04-29
Add `encoding.TextUnmarshaler` support for slice query parameters

B94715:
I must apologise, I accidentally pushed a PR (#1020) mistakenly commenting out tests that I did not own and didn't catch this until it was too late. I closed the PR immediately to rectify the issue but wasn't aware that GitHub automatically locks closed PRs when `--force-with-lease` is applied to a `git push`. The original PR was thus

--

This directly addresses #488  and is an alternative to #998.

`parseSliceInto()` was missing the encoding.TextUnmarshaler fallback that `parseInto()` already has. This aims to address that.


wolveix (approved):

wolveix:
Awesome work, thanks @B94715!

PR #1024, opened 2026-05-09
feat: support customizing the docs renderer config

johnletey:


wolveix:
Hey @johnletey! This is a cool contribution, thank you :)

I want to consider whether this may be usable by the other renderers too before considering merging. I'm a little limited on time right now, so if you happen to want to help with this, it'd be greatly appreciated!

johnletey:
Thanks @wolveix! Looked at both.

SwaggerUI was a good fit, so I've added it. Its config is just the `SwaggerUIBundle({...})` object, so `DocsRendererConfig` merges straight in. Only catch was CSP: the inline script is pinned by a sha256 hash, so I pass the config via a `data-config` attribute (same trick as Scalar) and read it at runtime. The script body stays static, so its hash stays valid.

Stoplight Elements doesn't fit the same shape though. It's configured via attributes on `<elements-api>` (router, layout, etc.), not one JSON object, so there's nowhere clean for an any to land. Left it out for now, but happy to do something dedicated if you'd prefer.

wolveix:
Awesome stuff, thank you so much @johnletey! Would you mind also adding/reworking the docs to demonstrate how these can be configured, please? I think all three docs examples can be simplified to utilize the `DocsRenderer` field, and could include a simple configurable map as an example with an `// Optional.` tag. Does that make sense to you?

here: https://github.com/johnletey/huma/blob/feat/scalar-docs-config/docs/docs/features/api-docs.md

johnletey:
@wolveix Is that what you had in mind?

wolveix:
Awesome stuff, thanks so much @johnletey!

PR #1038, opened 2026-05-31
fix(sse): flush response headers before user handler runs

### LINKED ISSUE #1037: SSE: response headers not flushed until first event, delaying EventSource.onopen
**Body:**

## Problem

`sse.Register` does not commit the 200 OK response headers until the first `send()` call writes event data. For SSE streams that wait for live events before sending anything, the browser's `EventSource` does not fire `onopen` until the first event arrives.

## Cause

1. `huma.Register` skips `ctx.SetStatus(status)` on the `outBodyFunc` path ([huma.go:1166-1168](https://github.com/danielgtaylor/huma/blob/main/huma.go#L1166-L1168)), so `WriteHeader(200)` is never called explicitly for streaming responses.

2. `sse.Register`'s `Body` callback sets `Content-Type` via `SetHeader` but never flushes before calling the user handler ([sse.go:133-215](https://github.com/danielgtaylor/huma/blob/main/sse/sse.go#L133-L215)). The first flush only happens inside `send()` after event data is written.

## Fix

The flusher discovery is moved up from the `send` closure (where it already exists) so the initial flush can use it. The duplicate discovery inside `send` can remain or be consolidated.

alswl:
## Summary

- `sse.Register` set `Content-Type: text/event-stream` but never committed the response status or flushed before invoking the user handler.
- For handlers that wait on a channel before sending the first event, the browser's `EventSource.onopen` did not fire until the first event arrived — headers stayed buffered.
- Fix: call `ctx.SetStatus(http.StatusOK)` and `flusher.Flush()` right after the flusher is discovered, so headers are pushed to the client immediately.

Fixes #1037

## Root cause

1. `huma.Register` skips `ctx.SetStatus(status)` on the `outBodyFunc` path ([huma.go:1166-1168](https://github.com/danielgtaylor/huma/blob/main/huma.go#L1166-L1168)), so `WriteHeader(200)` is never called for streaming responses.
2. `sse.Register`'s `Body` callback only flushed inside `send()` after writing event data, so until the first event the client saw no response headers.

## Why this matters — spec references

**[WHATWG HTML Living Standard §9.2 Server-sent events](https://html.spec.whatwg.org/multipage/server-sent-events.html)** binds `onopen` to the arrival of the response headers, not to the first event. Once the response is verified as `200` with `Content-Type: text/event-stream`, the user agent must *announce the connection*:

> When a user agent is to **announce the connection**, the user agent must queue a task which, if the `readyState` attribute is set to a value other than `CLOSED`, sets the `readyState` attribute to `OPEN` and **fires an event named `open`** at the `EventSource` object.

**[MDN — EventSource: open event](https://developer.mozilla.org/en-US/docs/Web/API/EventSource/open_event)** restates the same contract:

> The `open` event is fired when a connection with an event source is opened. This occurs **after the initial HTTP connection to the server has been successfully established**.

In Go's `net/http`, response headers are not put on the wire until `WriteHeader` is called *and* an `http.Flusher.Flush()` pushes the buffered bytes. Without both, a handler that blocks waiting for the first event leaves the client stuck in `CONNECTING` — violating the spec's expected timing.

## Test plan

- [x] `go test ./sse/... -count=1` — all existing tests pass
- [x] New regression test `sse flushes headers before first event` uses an `orderedWriter` to assert `WriteHeader(200)` and `Flush` both happen before the user handler blocks on a channel
- [x] Verified the new test fails on `main` without the fix and passes with it

### DIFF
--- a/sse/sse.go
+++ b/sse/sse.go
@@ -132,6 +132,9 @@ func Register[I any](api huma.API, op huma.Operation, eventTypeMap map[string]an
 		return &huma.StreamResponse{
 			Body: func(ctx huma.Context) {
 				ctx.SetHeader("Content-Type", "text/event-stream")
+				// Commit response headers immediately so the client's
+				// EventSource.onopen fires without waiting for the first event.
+				ctx.SetStatus(http.StatusOK)
 				bw := ctx.BodyWriter()
 				encoder := json.NewEncoder(bw)
 
@@ -148,6 +151,9 @@ func Register[I any](api huma.API, op huma.Operation, eventTypeMap map[string]an
 					} else {
 						break
 					}
+				}
+				if flusher != nil {
+					flusher.Flush()
 				}
 
 				var deadliner writeDeadliner
--- a/sse/sse_test.go
+++ b/sse/sse_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"slices"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


wolveix:
Awesome fix, thanks so much! This will be included in our next release.

PR #1045, opened 2026-06-12
fix: prevent panic in uniqueItems validation for unhashable types (#1042)

### LINKED ISSUE #1042: Validation of uniqueItems causes panic when input is not the expected type
Huma v2.38.0

If you have an input that is an array, and you use the uniqueItems validation, there are some paths that can lead to uncaught panics.

If the input is expecting an array of strings but you provide `"myArray": [{}]` this will trigger a panic.

Copilot took a look at Huma's codebase and came up with a theory that sounds cogent.
<details><summary>Copilot's analysis</summary>
<p>
Root cause: Huma v2's validator checks uniqueItems by hashing array elements before validating their types. When malformed input like [{}] arrives for a []string) field, the hasher panics on map[string]interface{}. Removing the tag prevents the hash attempt; huma still rejects [{}] with a proper 422 type-validation error since items must be strings.
</p>
<p>
(At this point I asked if it was a known issue)
</p>
<p>
No, it does not appear to be a reported issue. I searched the huma issue tracker for "uniqueItems panic", "unhashable", and "handleArray panic" — all returned zero results.

The bug is in handleArray in validate.go:
https://github.com/danielgtaylor/huma/blob/2b690545335bfc3ee2ae704109ec9c03480f9840/validate.go#L610

The uniqueness check inserts items into a map[any]struct{}, which panics if an item is a non-hashable type (like map[string]interface{}). This happens when malformed JSON (e.g. [{}] for a []string field) reaches uniqueness validation before type validation rejects it.

It would be worth filing upstream. The fix on their side would be to either defer uniqueItems checking until after item-level type validation, or use a recover / reflection-based comparison for non-hashable types.
</p>
</details> 

I've provided a minimal example to reproduce the issue at the bottom of the ticket.
Instructions / scenarios:

`curl -v -X POST -H 'Content-Type: application/json' -d '{"arrayOfStrings": ["one", "two"]}' http://localhost:8888/demo`
Expected: Returns 200 "OK"
Actual: Returns 200 "OK"

`curl -v -X POST -H 'Content-Type: application/json' -d '{"arrayOfStrings": ["one", "one"]}' http://localhost:8888/demo`
Expected: Returns 4xx because strings are not unique
Actual: Returns 422 because strings are not unique

`curl -v -X POST -H 'Content-Type: application/json' -d '{"arrayOfStrings": [{}]}' http://localhost:8888/demo`
Expected: Returns 4xx because array items are not strings
Actual: Server doesn't reply. Not even a 500.
Logs show: http: panic serving [::1]:59220: hash of unhashable type: map[string]interface {}

```go
// ./humatest/main.go
package main

import (
	"context"
	"fmt"
	"net/http"

	"github.com/go-chi/chi/v5"
)

type PostInput struct {
	Body struct {
		ArrayOfStrings []string `json:"arrayOfStrings" uniqueItems:"true"`
	}
}

type Output struct {
	Body string
}

func main() {
	r := chi.NewMux()
	api := humachi.New(r, huma.DefaultConfig("Bug Demo", "1.0.0"))

		Method: http.MethodPost,
		Path:   "/demo",
	}, func(ctx context.Context, input *PostInput) (*Output, error) {
		return &Output{Body: "OK"}, nil
	})

	fmt.Println("Listening on :8888")
	http.ListenAndServe(":8888", r)
}
```

If you're wondering how I found something this obscure, I was testing the API I was creating with Huma by using [Schemathesis](https://schemathesis.io/), which is a FOSS ([github repo](https://github.com/schemathesis/schemathesis)) API testing tools that I can't praise enough.

ArdyJunata:
## Summary

Fixes #1042

## Problem

The `uniqueItems` check in `handleArray` used `map[any]struct{}` as a hash set. When array items contain non-hashable types such as `map[string]interface{}` — produced by malformed JSON like `[{}]` for a `[]string` field — inserting into the map panics:

```
http: panic serving [::1]:59220: hash of unhashable type: map[string]interface {}
```

This caused the server to return **no response at all** instead of a `422 Unprocessable Entity`. The panic kills the goroutine serving the request.

**Reproduction:**
```bash
curl -X POST -H 'Content-Type: application/json' \
  -d '{"arrayOfStrings": [{}]}' \
  http://localhost:8888/demo
# Server logs: panic serving ...: hash of unhashable type: map[string]interface {}
```

## Fix

Introduce a `uniqueItemsContains` helper that:
1. **Fast path**: attempts an O(1) `map[any]` lookup (works for all hashable types — strings, ints, bools, etc.)
2. **Slow path**: a `deferred recover()` catches the panic for unhashable types and falls back to an O(n) `reflect.DeepEqual` linear scan

This means:
- ✅ Normal arrays of hashable items → same O(n) performance as before  
- ✅ Arrays with maps/slices → graceful O(n²) fallback, no panic  
- ✅ Duplicate unhashable objects → still detected correctly  
- ✅ Malformed `[{}]` for `[]string` → proper `422 expected string` error  

## Tests

Three regression tests added to `validate_test.go`:

| Test | Verifies |
|------|----------|
| `uniqueItems with unhashable object element must not panic` | `[{}]` for `[]string` returns `422`, not a panic |
| `uniqueItems detects duplicate unhashable objects` | `[{"a":1},{"a":1}]` is still detected as non-unique |
| `uniqueItems with mixed hashable and unhashable elements` | `["ok", {"k":"v"}]` does not panic |

All existing tests continue to pass.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -608,12 +608,41 @@ func handleArray[T any](r Registry, s *Schema, path *PathBuffer, mode ValidateMo
 	}
 
 	if s.UniqueItems {
-		seen := make(map[any]struct{}, len(arr))
+		// uniqueItemsContains reports whether item already appears in seen.
+		// It uses a map[any]struct{} for O(1) lookups but falls back to a
+		// reflect.DeepEqual linear scan when item is a non-hashable type
+		// (e.g. map[string]interface{} from malformed JSON like [{}]).
+		// Without this guard, inserting an unhashable key panics with
+		// "hash of unhashable type" and crashes the server (issue #1042).
+		uniqueItemsContains := func(seen []any, item any) (found bool) {
+			defer func() {
+				if recover() != nil {
+					// item is unhashable (e.g. a map or slice): fall back to
+					// deep equality so we still catch genuine duplicates.
+					for _, s := range seen {
+						if reflect.DeepEqual(s, item) {
+							found = true
+							return
+						}
+					}
+				}
+			}()
+			// Fast path: attempt a map lookup. This will panic for unhashable
+			// types, which the deferred recover above will catch.
+			m := make(map[any]struct{}, len(seen)+1)
+			for _, s := range seen {
+				m[s] = struct{}{}
+			}
+			_, found = m[item]
+			return
+		}
+
+		seen := make([]any, 0, len(arr))
 		for _, item := range arr {
-			if _, ok := seen[item]; ok {
+			if uniqueItemsContains(seen, item) {
 				res.Add(path, arr, validation.MsgExpectedArrayItemsUnique)
 			}
-			seen[item] = struct{}{}
+			seen = append(seen, item)
 		}
 	}
 
--- a/validate.go
+++ b/validate.go
@@ -608,41 +608,48 @@ func handleArray[T any](r Registry, s *Schema, path *PathBuffer, mode ValidateMo
 	}
 
 	if s.UniqueItems {
-		// uniqueItemsContains reports whether item already appears in seen.
-		// It uses a map[any]struct{} for O(1) lookups but falls back to a
-		// reflect.DeepEqual linear scan when item is a non-hashable type
-		// (e.g. map[string]interface{} from malformed JSON like [{}]).
-		// Without this guard, inserting an unhashable key panics with
-		// "hash of unhashable type" and crashes the server (issue #1042).
-		uniqueItemsContains := func(seen []any, item any) (found bool) {
-			defer func() {
-				if recover() != nil {
-					// item is unhashable (e.g. a map or slice): fall back to
-					// deep equality so we still catch genuine duplicates.
-					for _, s := range seen {
-						if reflect.DeepEqual(s, item) {
-							found = true
-							return
-						}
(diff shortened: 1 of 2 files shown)


wolveix:
Thanks for tracking this down, @ArdyJunata! The panic fix is correct and the three regression tests cover the right cases.

I pushed a performance follow-up directly to your branch. `uniqueItemsContains` rebuilt a fresh `map[any]` from all previously-seen items on every element, which made the check O(n²) for every uniqueItems array, not only the unhashable case.

Benchmarks before and after the changes:
  
  ```md
┌───────┬────────────────┬──────────────┬────────────────────┐
  │ items │ before this PR │ with this PR │ with the follow-up │
  ├───────┼────────────────┼──────────────┼────────────────────┤
  │ 1,000 │ 0.03 ms        │ 8.5 ms       │ 0.03 ms            │
  ├───────┼────────────────┼──────────────┼────────────────────┤
  │ 4,000 │ 0.10 ms        │ 149 ms       │ 0.13 ms            │
  └───────┴────────────────┴──────────────┴────────────────────┘
```

The follow-up keeps your recover approach; it grows a single map incrementally for hashable items (O(1) lookups, the common case) and only falls back to the reflect.DeepEqual scan for genuinely unhashable items.

PR #1052, opened 2026-06-29
fix: content type validation should be case insensitive

hlavacekvit:
I ran into this in production: an old .NET app couldn't call our Huma-based service because it sends Content-Type: Application/Json (capitalized), and Huma responded with 415 Unsupported Media Type.

Media types are case-insensitive per RFC 9110 §8.3.1, but the format lookup (a.formats[ct]) is a case-sensitive map access, so Application/Json doesn't match the registered application/json.

Fixed by lowercasing the media-type token. Added a test as well

### DIFF
--- a/api.go
+++ b/api.go
@@ -352,7 +352,7 @@ func (a *api) Unmarshal(contentType string, data []byte, v any) error {
 		return err
 	}
 
-	ct := contentType[start:end]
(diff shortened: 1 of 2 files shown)


wolveix:
Awesome find and fix, thanks @hlavacekvit!

Issue #286, opened 2024-03-06
Example for handling long lived SSE connections


IcelandicIcecream:
Hey! Great work on the framework, gonna use it for most of my projects from now on. Just wanted to ask if there is an example for a long lived SSE connection. Like how do I handle user disconnects and etc. Also, how do I handle errors for SSE connections? Im using the example with send.Data as reference

Insei:
You can check that you client is alive by send.Data(), this func return error if connection lost or connection has any other error. For example you can send ping event every 10s. Yep, this is not perfect solution, but it can works correctly.

IcelandicIcecream:
> Hi! You can check that you client is alive by send.Data(), this func return error if connection lost or connection has any other error. For example you can send ping event every 10s. Yep, this is not perfect solution, but it can works correctly.

Alright. thanks for the suggestion! Will do this in the meantime. If I remember correctly, maybe we can actually run a separate goroutine to check for ctx.Done(). Will do some testing and let you know!

danielgtaylor:
@IcelandicIcecream FYI there is a more complete FizzBuzz SSE example here:

https://github.com/danielgtaylor/huma/blob/main/examples/sse/main.go

It includes ending the handler when `send.Data(...)` fails:

https://github.com/danielgtaylor/huma/blob/main/examples/sse/main.go#L156-L158

danielgtaylor:
Please let me know if the provided examples don't answer your question, and feel free to re-open if needed!

artiga033:
There are also [comments](https://html.spec.whatwg.org/multipage/server-sent-events.html#:~:text=comment%20%20%20%20%20%20%20%3D%20colon%20*any%2Dchar%20end%2Dof%2Dline) defined in the standard. [MDN also suggests](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#:~:text=Note%3A%20The%20comment%20line%20can%20be%20used%20to%20prevent%20connections%20from%20timing%20out%3B%20a%20server%20can%20send%20a%20comment%20periodically%20to%20keep%20the%20connection%20alive.) using this for keepalive. As far as I see huma only supports sending events.

sjayach:
@danielgtaylor does HUMA support comments?

PR #1054, opened 2026-07-10
Support sending comments with SSE

seanwu1105:
Sending comments is a common way to keep the connection alive for SSE. This PR adds the feature and refactor to reuse `flush` util function.

This potentially address the questions in https://github.com/danielgtaylor/huma/issues/286.


wolveix:
Thanks so much @seanwu1105!

PR #1060, opened 2026-07-12
feat(form-data): handle unmarshalling and validation of non-file JSON form data fields

### LINKED ISSUE #897: Parsing a struct in `Multipart Form Data` panics
I've seen that parsing arbitrary types has been implemented just recently.
```sh
curl http://localhost:9999/foo \
    --form 'name=josh' \
    --form 'content=@abc.pdf'
```
Parses fine into:
```go
type Input struct {
	RawBody huma.MultipartFormFiles[struct {
		Name    string        `form:"name"`
		Content huma.FormFile `form:"content"`
	}]
}
```

---
But imagine a scenario where I need to upload a file (possibly binary) to the server. And alongside this file, I want to send some additional metadata as JSON. Like this:
```sh
curl http://localhost:9999/foo \
    --form 'metadata={\"client\":\"curl\", \"user_id\":1}' \
    --form 'content=@abc.pdf'
```
Then, I would try to define the Input struct as follows:
```go
type Input struct {
	RawBody huma.MultipartFormFiles[struct {
		Metadata Metadata      `form:"metadata" contentType:"application/json"`
		Content  huma.FormFile `form:"content"`
	}]
}
type Metadata struct {
	Client string `json:"client"`
	UserID int    `json:"user_id"`
}
```
With that, Huma even generates the right OpenAPI spec:
```yaml
...
Metadata:
  additionalProperties: false
  properties:
    client:
      type: string
    user_id:
      format: int64
      type: integer
  required:
    - client
    - user_id
  type: object
...
requestBody:
  content:
    multipart/form-data:
      encoding:
        content:
          contentType: application/octet-stream
        metadata:
          contentType: text/plain
      schema:
        properties:
          content:
            contentEncoding: binary
            contentMediaType: application/octet-stream
            format: binary
            type: string
          metadata:
            $ref: "#/components/schemas/Metadata"
        type: object
...
```
But sadly, the parsing then fails when a request is fired.
```
http: panic serving [::1]:50273: unsupported param type Metadata
```

lsdch:
Closes #897.

Makes Huma unmarshal and validate non-file fields in multipart forms that have tag `contentType:"application/json"`.

I am not perfectly happy with this implementation, since it duplicates a bit of logic from regular JSON body unmarshalling. But factorizing this would involve major changes in the API.

To keep the patch minimal, I had to include the parsed `contentType` tag in `fieldParamInfo`. Hope this is acceptable.

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -181,6 +181,51 @@ type ComplexOutput struct {
 	}
 }
 
+func BenchmarkMultipartJSONField(b *testing.B) {
+	_, api := humatest.New(b, huma.DefaultConfig("Test API", "1.0.0"))
+
+	type Metadata struct {
+		Name   string `json:"name" default:"anon"`
+		Email  string `json:"email" format:"email"`
+		Age    int    `json:"age" default:"18"`
+		Client string `json:"client"`
+	}
+
+	huma.Register(api, huma.Operation{
+		Method: http.MethodPost,
+		Path:   "/upload",
+	}, func(ctx context.Context, input *struct {
+		RawBody huma.MultipartFormFiles[struct {
+			Metadata Metadata      `form:"metadata" contentType:"application/json" required:"true"`
+			File     huma.FormFile `form:"file" contentType:"text/plain"`
+		}]
+	}) (*struct{}, error) {
+		return nil, nil
+	})
+
+	body := "--SimpleBoundary\r\n" +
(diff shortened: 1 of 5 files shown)


wolveix:
Thanks @lsdch!

lsdch:
And thank you for improving the PR !

PR #1065, opened 2026-07-17
fix: do not panic on unresolved schema $ref during Validate

### LINKED ISSUE #1063: [Bug]: Validate panics on unresolved schema $ref instead of validation error
### Describe the bug

`Validate` follows schema `$ref`s via `Registry.SchemaFromRef` without a nil check. If a ref is missing/unresolved, `SchemaFromRef` returns `nil` and the next field access panics:

```go
for s.Ref != "" {
    s = r.SchemaFromRef(s.Ref)
}
// s may be nil → panic on use
```

Same pattern appears for object properties, array items, and `allOf` subschemas.

### Debug information

- Checkout @ `d6f2a37d705e0b8159bd5cc92f7681b1d7d4778e`
- Files: `validate.go`, `registry.go`

### Repro steps

Build a registry/schema with a `$ref` that does not resolve (missing component), then call `Validate` on a request body that walks that ref.

Observed: `runtime error: invalid memory address or nil pointer dereference` for top-level, property, items, and allOf missing refs.

### Expected behavior

Unresolved `$ref` should produce a validation / schema error (e.g. 422-style `ValidateResult` entry), not a process panic.

### Actual behavior

Nil pointer panic on normal request validation when the schema graph is incomplete.

### Suggested direction

After each `SchemaFromRef`, if `s == nil`, add an “unresolved $ref” validation error and return; never dereference nil.

I'd be happy to send a PR with tests.

baiyuxi930826:
## Summary

Fixes #1063: `Validate` no longer panics when a schema `$ref` cannot be resolved.

`SchemaFromRef` can return nil for missing keys; the previous loops dereferenced the result without a nil check. Unresolved refs now add a validation error (`expected schema $ref to resolve: ...`) and return.

Covered sites:
- top-level ref resolution in `Validate`
- discriminator mapping refs
- property refs in `handleMapString` / `handleMapAny`

## Test plan

- [x] `go test . -run TestValidateUnresolvedSchemaRefNoPanic`
- [x] Confirms top-level and property missing refs produce errors without panic


wolveix:
Thanks for this. From reviewing, it's not really possible in a normal request to ever hit this; however, it's better to be defensive when it's minimal to do so.

PR #1079, opened 2026-07-28
fix: Ensure autopatch processes properties in the same order every time

lucaspopp-wbd:
Since map iterators return properties in a random order, autopatch schemas can fluctuate when multiple fields use the same type.

This change ensures fields are processed in a deterministic order (sorted by name) to avoid type name fluctuations.

### DIFF
--- a/autopatch/autopatch.go
+++ b/autopatch/autopatch.go
@@ -10,14 +10,17 @@ package autopatch
 
 import (
 	"bytes"
+	"cmp"
 	"context"
 	"encoding/json"
 	"errors"
 	"io"
+	"iter"
 	"maps"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


wolveix:
Thanks for the PR, but I don't think this change can affect the generated schema.

`makeOptionalSchema` is side-effect free, and its only action here is inserting into `optionalSchema.Properties`, where map insertion order isn't observable. `Schema.MarshalJSON` emits `properties` as a `map[string]*Schema`, so `encoding/json` already sorts the keys, and nothing in this path touches the registry. I dumped the spec 500 times on `22a3cb0` with several same-typed inline fields across four autopatched resources: one distinct output, keys already sorted.

The tests only exercise `sortedMapIterator` itself, so they'd pass with `makeOptionalSchema` untouched. Could you add a test that reproduces the fluctuation, i.e. one that fails on `main`, or share the setup where names fluctuate?

Two things that may be closer to what you're seeing:

1. `makeOptionalSchema` doesn't copy `s.Ref`, so every `$ref` property (any nested struct field) collapses to `{}` in the merge-patch body.
2. `for _, path := range oapi.Paths` in `AutoPatch` is real map-order iteration, so PATCH operations register in random order.

If we do keep the sort, I'd inline it rather than add a helper, matching `schema.go`:

```go
for _, k := range slices.Sorted(maps.Keys(s.Properties)) {
	optionalSchema.Properties[k] = makeOptionalSchema(s.Properties[k])
}
```

lucaspopp-wbd:
Good point! Closed the PR :)

PR #1080, opened 2026-07-28
fix: Allow autopatch to resolve refs, when making the put schema optional

lucaspopp-wbd:
Pass the schema registry to autopatch's `makeOptionalSchema` method, so it can be leveraged to resolve refs

### DIFF
--- a/autopatch/autopatch.go
+++ b/autopatch/autopatch.go
@@ -236,7 +236,7 @@ func PatchResource(api huma.API, path *huma.PathItem) {
 	putSchema := put.RequestBody.Content["application/json"].Schema
 
 	// Create an optional version of the PUT schema
-	optionalPutSchema := makeOptionalSchema(api.OpenAPI().Components.Schemas, putSchema)
+	optionalPutSchema := makeOptionalSchema(api.OpenAPI().Components.Schemas, putSchema, map[string]struct{}{})
 
 	// Manually register the operation so it shows up in the generated OpenAPI.
 	op := &huma.Operation{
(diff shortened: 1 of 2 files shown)


wolveix:
Thanks for this. Two blockers before merge, both reproducible.

**Stack overflow on recursive schemas.** The deref at `autopatch.go:458` is unconditional, so a self-referential type never terminates. This passes on `22a3cb0` and crashes at startup here (the old code survived only because it dropped refs):

```go
type Node struct {
	Name     string  `json:"name"`
	Children []*Node `json:"children,omitempty"`
}
// register GET + PUT for Node, then AutoPatch(api) => fatal error: stack overflow
```

**Nil panic on unresolvable refs.** `SchemaFromRef` returns nil when the ref doesn't match the registry prefix (`registry.go:174`), so `s` goes nil and the literal at `:462` panics:

```go
makeOptionalSchema(registry, &huma.Schema{Ref: "https://example.com/schemas/Thing.json"})
```

A visited set handles both, matching how `schema.go:702` and `huma.go:556` already thread cycle guards:

```go
if s.Ref != "" {
	resolved := registry.SchemaFromRef(s.Ref)
	if _, cycle := visited[s.Ref]; cycle || resolved == nil {
		// Unresolvable or self-referential: leave the ref in place.
		return &huma.Schema{Ref: s.Ref}
	}
	visited[s.Ref] = struct{}{}
	defer delete(visited, s.Ref)
	s = resolved
}
```

Keeping the ref on a cycle leaves nested `required` intact in that one spot, which beats crashing. Please add tests for both cases.

Minor:

- `:239` uses `api.OpenAPI().Components.Schemas` when `oapi` is already in scope.
- `AdditionalProperties` is `any` and copied verbatim, so a `*Schema` there keeps its refs and its `required`.

lucaspopp-wbd:
@wolveix all great points! Addressed all of your feedback

lucaspopp-wbd:
Hey @wolveix wondering if you can take another look at this soon?

Would love to get this merged in so my team's SDK can leverage this!

PR #1115, opened 2026-09-23
perf: keep Register's generic part small to cut compile time

iyuneev:
`Register[I, O]` is one large generic function, so Go compiles a separate copy of the whole body for every distinct input/output pair. Only a couple of lines actually need `I` and `O` (creating the input and calling the handler); everything else already goes through reflection.

This moves the body into a non-generic `registerCore` and keeps `Register` as a thin generic wrapper. The public API doesn't change and no call sites need to be touched.

### Numbers

A package with 200 operations, each with its own input/output types, rebuilding just that package (Go 1.27, median of 5):

| | main | this PR |
|---|---:|---:|
| build time | 3.38 s | 0.64 s |
| object code from `Register` instantiations | 4,553 KiB | 200 KiB |
| package archive | 33.1 MB | 6.8 MB |

In our own API (~640 operations across 27 packages) this took `Register` from ~19.7 KB to ~0.8 KB of object code per operation, and cold compile CPU time for those packages from 71.5 s to 23.1 s.

Runtime is unaffected. benchstat over 12 interleaved runs shows no significant difference for `FullRequest_Complex`, `HandlerFunc` or `Register`, and per-request allocations are identical. Registration itself does 2 extra allocations (the two closures).

### One thing worth a look

The handler returns `*O`, and a nil `*O` stored in an `any` is not `== nil`. The wrapper returns an untyped nil in that case so the existing "no output" path still triggers; `TestFeatures/response-nil` covers it.

`go test ./...` passes, apart from `yaml.TestConvert` and `humacli.TestCLIShutdown`, which fail the same way on `main` on my machine (Windows, Go 1.27).


wolveix:
Really fantastic performance find @iyuneev! In one of my projects, I found a 7.3% binary size reduction (46.4MB -> 43MB), build time reduction from 15s -> 14s, but our project has 141 register calls compared to your 640 :) Great work!

